PR #43, opened 2022-10-13
Ci/merge workflows into one

### LINKED ISSUE #42: Update CI workflow
Currently, the repo has two workflows (`linters.yml` and `tests.yml`). We can merge those workflows into one workflow with two jobs. The linter job must execute in a Docker container (`urnai/linter:latest`). The new workflow must be:

```mermaid
  graph LR;
      L[Lint]-->T[Tests];
```

sudiptob2:
closes #42 

- [x] merge 2 workflows (lint and test)
- [x] test depends on linting
- [x] linting runs on docker container


sudiptob2:
@alvarofpp could you please review the PR :pray: #42 
Also, I am a little bit confused regarding what command I should put in the `urnai/linter:latest` :confused: 
Currently, I am only running `lint-commit` as you can see. Please suggest any edits :blush:

alvarofpp:
@sudiptob2 You have to set it to execute 3 commands that can be seen [here](https://github.com/UFRN-URNAI/urnai-tools/blob/dca2dd18cbfbbc7e6c2ea730d167c2f62410ef63/Makefile#L29).

.github/workflows/ci.yml
```diff
@@ -8,8 +8,16 @@ on:
     branches: [ master ]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    container: urnai/linter:latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: lint-commit
```
alvarofpp, line 16: `lint-commit` needs a target branch as a parameter. The correct value is `origin/master`.

alvarofpp (changes requested):

.github/workflows/ci.yml
sudiptob2: understood

alvarofpp:
@sudiptob2 There were some errors in the linter job, could you fix them? To test it locally, you can run `make lint` on the repository (you need the Makefile on your machine).

sudiptob2:
instead of running like 
```bash
    runs-on: ubuntu-latest
    container: urnai/linter:latest
    steps:
      - uses: actions/checkout@v3
      - run: lint-commit origin/master
      - run: lint-markdown
      - run: lint-python urnai
```
can we just run lint like this as before ---
```
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - run: make lint
 ```
@alvarofpp

alvarofpp:
@sudiptob2 The problem is that we have a `docker pull` command in `make lint`, so it would be a container within a container for lint (and this is bad practice). To avoid blocking your contribution, remove the line `container: urnai/linter:latest` and add `make lint` and it should work.

alvarofpp:
@sudiptob2 I don't see a problem with this approach. Running the `urnai/linter:latest` container can be left for another issue.

sudiptob2:
Thanks, I updated the code. Running directly in the container is better though, I would like to investigate that also :pray:

alvarofpp:
@sudiptob2 The linter job had this error [here](https://github.com/UFRN-URNAI/urnai-tools/actions/runs/3245414969/jobs/5323285432#step:3:62), but the pipeline finished successfully. I created card #44 to investigate further and fix this error in the linter job. Thank you very much for your contribution!

PR #46, opened 2022-10-14
Ci/fix lint job

### LINKED ISSUE #44: Fix linter job
The `make lint` command works locally, but when it is run in the `ci` pipeline it throws an error (see [here](https://github.com/UFRN-URNAI/urnai-tools/actions/runs/3245414969/jobs/5323285432#step:3:62)) and doesn't check anything. This must be corrected so that the linter can execute its commands correctly.

An extra is to have this job run in a container from the `urnai/linter:latest` image.

sudiptob2:
fixes #44 

### Observations
Due to the fact that GitHub action does a shallow clone of the repo (using `actions/checkout@v3`) the checked-out repository does not have a branch `master`. As a result, it gives `fatal: ambiguous argument 'origin/master'` error when `make lint` is run.

### Fix
By forcing `fetch-depth: 0` we can checkout the entire repo and the `origin/master` will be available. :blush: 

### Opinion
I am not getting the point of running `make lint` on `master` branch because the code that needs to be checked by the linter are in the current branch. Shouldn't we run the workflow on the `origin/current-PR-branch`? :pray:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,6 @@
 name: Linting and Unit-test
 
 on:
-  push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
 
@@ -12,6 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - run: make lint
 
   test:


sudiptob2:
@alvarofpp can you please review :pray:

alvarofpp:
@sudiptob2 Could you use only a single commit in this PR? In this case, it would remove commits `15b138c` and `9a31b8b`.

> Shouldn't we run the workflow on the origin/current-PR-branch?

I agree with you. Since this is an integration workflow, I think you could remove the `push: branches: [ master ]`.

alvarofpp (changes requested):
@sudiptob2 Could you use only a single commit in this PR? In this case, it would remove commits `15b138c` and `9a31b8b`.

> Shouldn't we run the workflow on the origin/current-PR-branch?

I agree with you. Since this is an integration workflow, I think you could remove the `push: branches: [ master ]`.

alvarofpp:
@sudiptob2 The `push` event will trigger the jobs when a branch is pushed, regardless of which branch it is (the `master` or the one someone else is working on). The intention would be to trigger only during a Pull Request, that's why I think the `pull_request` event is ideal, besides that it occurs, by default, in 3 types of activities (`opened`, `synchronize`, and `reopened`). So we would have:

```yml
on:
  pull_request:
    branches: [ master ]
```

sudiptob2:
Thanks understood :pray:

alvarofpp (approved):

alvarofpp:
@sudiptob2 Thank you for your contribution!

PR #60, opened 2023-06-21
Add code of conduct #9

thuthu-c:
Added "CODE_OF_CONDUCT.md" file.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,0 +1,133 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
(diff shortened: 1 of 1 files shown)


alvarofpp (changes requested):
Fixes:

- It is necessary that you follow the pattern of Conventional Commits in your commits.
- Commits must be clear and objective. In your case, it does not make sense to have a commit that creates and updates the same file (if it did not exist in the repository, then it is only an add and not an update).
- You can summarize everything in a single commit, there is no need for 2 commits.
- Need to improve the title of the PR.

Suggestion:

- Add your references for that code of conduct in the PR.

PR #78, opened 2024-02-27
Resolves #72

### LINKED ISSUE #72: Base class for the state representation
We need a class that lets us know the agent's current state. It needs to have the `update` method, which will update the agent's state according to the script developed.

Suggestions:

- Class name: `StateBase`.
- Class path: `urnai.states`.

RickFqt:
- Resolves #72.

### DIFF
--- a/tests/units/states/test_state_base.py
+++ b/tests/units/states/test_state_base.py
@@ -13,11 +13,11 @@ class TestStateBase(unittest.TestCase):
             def __init__(self):
                 super().__init__()
 
-        f = FakeState()
-        update_return = f.update("observation")
-        state = f.get_state
-        dimension = f.get_dimension
-        reset_return = f.reset()
+        fake_state = FakeState()
+        update_return = fake_state.update("observation")
+        state = fake_state.state
+        dimension = fake_state.dimension
+        reset_return = fake_state.reset()
         assert isinstance(StateBase, ABCMeta)
         assert update_return is None
         assert state is None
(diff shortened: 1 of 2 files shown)


tests/units/states/test_state_base.py
```diff
@@ -0,0 +1,25 @@
...
+        StateBase.__abstractmethods__ = set()
+
+        class FakeState(StateBase):
+            def __init__(self):
+                super().__init__()
+
+        f = FakeState()
+        update_return = f.update("observation")
+        state = f.get_state
+        dimension = f.get_dimension
```
alvarofpp: Ideally, it shouldn't be `get_<something>`, but just the name of the attribute. Example: instead of `f.get_state` it would be `f.state`.

tests/units/states/test_state_base.py
```diff
@@ -0,0 +1,25 @@
...
+class TestStateBase(unittest.TestCase):
+
+    def test_abstract_methods(self):
+        StateBase.__abstractmethods__ = set()
+
+        class FakeState(StateBase):
+            def __init__(self):
+                super().__init__()
+
+        f = FakeState()
```
alvarofpp: Better would be `fake_state` instead of just `f`. This was in the last PR (and will be fixed in the future), so we should be careful not to repeat it.

alvarofpp (changes requested):

alvarofpp (approved):

PR #79, opened 2024-03-01
Resolves #67

### LINKED ISSUE #67: Create base classes for Action
We should have a base class for developing agent actions, these would be the micro-actions. In addition, we need a class for collecting actions, these would be macro-actions. These classes need to have methods that make it possible to know whether or not it is possible to perform that action.

Suggestions:

- Class names: `ActionBase` (micro-action) and `ChainOfActionsBase` (macro-actions).
- Class paths: `urnai.actions`.

Notes:

CinquilCinquil:
#67

### DIFF
--- a/tests/units/actions/test_action_base.py
+++ b/tests/units/actions/test_action_base.py
@@ -12,10 +12,10 @@ class TestActionBase(unittest.TestCase):
         class FakeAction(ActionBase):
             ...
 
-        f = FakeAction()
-        run_return = f.run()
-        check_return = f.check("observation")
-        is_complete_return = f.is_complete()
+        fake_action = FakeAction()
+        run_return = fake_action.run()
+        check_return = fake_action.check("observation")
+        is_complete_return = fake_action.is_complete()
         assert isinstance(ActionBase, ABCMeta)
         assert run_return is None
         assert check_return is None
--- a/tests/units/actions/test_action_base.py
+++ b/tests/units/actions/test_action_base.py
@@ -4,13 +4,13 @@ from abc import ABCMeta
 from urnai.actions.action_base import ActionBase
 
 
+class FakeAction(ActionBase):
+        ...
+
 class TestActionBase(unittest.TestCase):
 
     def test_abstract_methods(self):
         ActionBase.__abstractmethods__ = set()
-
-        class FakeAction(ActionBase):
-            ...
 
         fake_action = FakeAction()
         run_return = fake_action.run()
--- a/tests/units/actions/test_action_base.py
+++ b/tests/units/actions/test_action_base.py
@@ -5,18 +5,21 @@ from urnai.actions.action_base import ActionBase
 
 
 class FakeAction(ActionBase):
-        ...
+    ActionBase.__abstractmethods__ = set()
+    __id__ = None
+    ...
 
 class TestActionBase(unittest.TestCase):
 
     def test_abstract_methods(self):
-        ActionBase.__abstractmethods__ = set()
 
         fake_action = FakeAction()
         run_return = fake_action.run()
         check_return = fake_action.check("observation")
         is_complete_return = fake_action.is_complete()
+        assert fake_action.__id__ is None
         assert isinstance(ActionBase, ABCMeta)
         assert run_return is None
         assert check_return is None
         assert is_complete_return is None
+
--- a/tests/units/actions/test_action_base.py
+++ b/tests/units/actions/test_action_base.py
@@ -18,5 +18,5 @@ class TestActionBase(unittest.TestCase):
         is_complete_return = f.is_complete()
         assert isinstance(ActionBase, ABCMeta)
         assert run_return is None
-        assert check_return is False
-        assert is_complete_return is False
+        assert check_return is None
+        assert is_complete_return is None
--- a/tests/units/actions/test_chain_of_actions_base.py
+++ b/tests/units/actions/test_chain_of_actions_base.py
@@ -16,9 +16,7 @@ class TestChainOfActionsBase(unittest.TestCase):
 
 		fake_chain_of_actions = FakeChainOfActions()
 		get_action_return = fake_chain_of_actions.get_action(0)
-		check_return = fake_chain_of_actions.check()
 		length_return = fake_chain_of_actions.length
 		assert isinstance(ChainOfActionsBase, ABCMeta)
 		assert get_action_return is None
-		assert check_return is True
 		assert (length_return == 0) is True
--- a/tests/units/actions/test_chain_of_actions_base.py
+++ b/tests/units/actions/test_chain_of_actions_base.py
@@ -5,8 +5,8 @@ from urnai.actions.chain_of_actions_base import ChainOfActionsBase
 
 
 class FakeChainOfActions(ChainOfActionsBase):
-	def __init__(self):
-		super().__init__()
+		def __init__(self):
+			super().__init__()
 
 class TestChainOfActionsBase(unittest.TestCase):
 
--- a/tests/units/actions/test_chain_of_actions_base.py
+++ b/tests/units/actions/test_chain_of_actions_base.py
@@ -5,13 +5,14 @@ from urnai.actions.chain_of_actions_base import ChainOfActionsBase
 
 
 class FakeChainOfActions(ChainOfActionsBase):
-		def __init__(self):
-			super().__init__()
+	ChainOfActionsBase.__abstractmethods__ = set()
+	
+	def __init__(self):
+		super().__init__()
 
 class TestChainOfActionsBase(unittest.TestCase):
 
 	def test_abstract_methods(self):
-		ChainOfActionsBase.__abstractmethods__ = set()
 
 		fake_chain_of_actions = FakeChainOfActions()
 		get_action_return = fake_chain_of_actions.get_action(0)
--- a/tests/units/actions/test_chain_of_actions_base.py
+++ b/tests/units/actions/test_chain_of_actions_base.py
@@ -4,20 +4,20 @@ from abc import ABCMeta
 from urnai.actions.chain_of_actions_base import ChainOfActionsBase
 
 
+class FakeChainOfActions(ChainOfActionsBase):
+	def __init__(self):
+		super().__init__()
+
 class TestChainOfActionsBase(unittest.TestCase):
 
 	def test_abstract_methods(self):
 		ChainOfActionsBase.__abstractmethods__ = set()
 
-		class FakeChainOfActions(ChainOfActionsBase):
-			def __init__(self):
-				super().__init__()
-
-		f = FakeChainOfActions()
-		get_action_return = f.get_action(0)
-		check_return = f.check()
-		len_return = f.len
+		fake_chain_of_actions = FakeChainOfActions()
+		get_action_return = fake_chain_of_actions.get_action(0)
+		check_return = fake_chain_of_actions.check()
+		length_return = fake_chain_of_actions.length
 		assert isinstance(ChainOfActionsBase, ABCMeta)
 		assert get_action_return is None
 		assert check_return is True
-		assert (len_return == 0) is True
+		assert (length_return == 0) is True
--- a/tests/units/actions/test_chain_of_actions_base.py
+++ b/tests/units/actions/test_chain_of_actions_base.py
@@ -1,0 +1,23 @@
+import unittest
+from abc import ABCMeta
+
+from urnai.actions.chain_of_actions_base import ChainOfActionsBase
+
+
+class TestChainOfActionsBase(unittest.TestCase):
+
+	def test_abstract_methods(self):
+		ChainOfActionsBase.__abstractmethods__ = set()
+
+		class FakeChainOfActions(ChainOfActionsBase):
+			def __init__(self):
+				super().__init__()
+
+		f = FakeChainOfActions()
+		get_action_return = f.get_action(0)
+		check_return = f.check()
+		len_return = f.len
+		assert isinstance(ChainOfActionsBase, ABCMeta)
+		assert get_action_return is None
+		assert check_return is True
+		assert (len_return == 0) is True
--- a/urnai/actions/action_base.py
+++ b/urnai/actions/action_base.py
@@ -2,13 +2,11 @@ from abc import ABC, abstractmethod
 
 
 class ActionBase(ABC):
-	
-	def __init__(self, id):
-		self.id = id
+	id = None
 	
 	@abstractmethod
 	def run(self):
-		"""Contains logic for executing the action"""
+		"""Executing the action"""
 		...
 	
 	@abstractmethod
@@ -20,6 +18,3 @@ class ActionBase(ABC):
 	def is_complete(self) -> bool:
 		"""Returns whether the action has finished or not"""
 		...
-		
-	
-
--- a/urnai/actions/action_base.py
+++ b/urnai/actions/action_base.py
@@ -1,25 +1,25 @@
 from abc import ABC, abstractmethod
 
-from urnai.utils.returns import ActionIndex, List
-
 
 class ActionBase(ABC):
+	
+	def __init__(self, id):
+		self.id = id
+	
+	@abstractmethod
+	def run(self):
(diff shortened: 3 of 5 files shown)


urnai/actions/action_base.py
```diff
@@ -0,0 +1,25 @@
+from abc import ABC, abstractmethod
+
+
+class ActionBase(ABC):
+	
+	def __init__(self, id):
+		self.id = id
```
alvarofpp: I believe that the identifier will not be passed dynamically during the initialization of an object, but will be declared when the class is developed. Therefore, I think it would look better like this:

```py
class ActionBase(ABC):
    id = None
```

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		...
+	
+	@abstractmethod
+	def get_first(self) -> ActionBase:
+		"""Returns the first action in the sequence"""
+		...
+	
+	@abstractmethod
+	def add_action(self, action: ActionBase):
+		"""Contains logic for adding an action to the sequence"""
```
alvarofpp: "Contains logic" is redundant, because every function contains the logic of something.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		...
+		
+	@abstractmethod
+	def remove_action(self, action: ActionBase):
+		"""Contains logic for removing an action from the sequence"""
```
alvarofpp: "Contains logic" is redundant, because every function contains the logic of something.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		...
+		
+	@abstractmethod
+	def clear():
+		...
+	
+	@abstractmethod
+	def merge(self, chain_of_actions):
+		"""Contains logic for merging a list of sequences of actions into self"""
```
alvarofpp: "Contains logic" is redundant, because every function contains the logic of something.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		...
+	
+	# Executing the next action in the sequence
+	def next(self):
```
alvarofpp: I think the function comment should be below and not above the function signature.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		first_action = self.get_first()
+		
+		if first_action.is_complete():
+			self.remove_action(first_action)
+			self.next()
+		else:
+			first_action.run()
+	
+	# Checking all of the actions in the sequence
+	def check(self) -> bool:
```
alvarofpp: I think the function comment should be below and not above the function signature.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		actions = self.get_actions()
+		return all(a.check() for a in actions)
```
alvarofpp:
```suggestion
		return all(action.check() for action in actions)
```

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+
+	def get_len(self):
```
alvarofpp:
```suggestion
	def get_len(self) -> int:
```

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		"""Contains logic for adding an action to the sequence"""
+		...
+		
+	@abstractmethod
+	def remove_action(self, action: ActionBase):
+		"""Contains logic for removing an action from the sequence"""
+		...
+		
+	@abstractmethod
+	def clear():
```
alvarofpp: The function description is missing.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+	
+	# Executing the next action in the sequence
+	def next(self):
+		if (self.get_len() == 0): 
+			return None
+	
+		first_action = self.get_first()
+		
+		if first_action.is_complete():
+			self.remove_action(first_action)
```
alvarofpp: If you're always going to remove the first element, I think something like `list.pop()` would be more appropriate.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
...
+		...
+	
+	@abstractmethod
+	def merge(self, chain_of_actions):
+		"""Contains logic for merging a list of sequences of actions into self"""
+		...
+	
+	# Executing the next action in the sequence
+	def next(self):
+		if (self.get_len() == 0):
```
alvarofpp:
```suggestion
		if self.get_len() == 0:
```

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,60 @@
+from abc import ABC, abstractmethod
+
+from action_base import ActionBase
+
+from urnai.utils.returns import List
+
+
+class ChainOfActionsBase(ABC):
```
alvarofpp, line 6: I couldn't identify the variable that will contain the actions. Also, methods like `get_actions`, `remove_action`, and others can be replaced by this variable that will contain the actions.

urnai/actions/action_base.py
```diff
@@ -0,0 +1,25 @@
...
+		"""Returns whether the action can be executed or not"""
+		...
+	
+	@abstractmethod
+	def is_complete(self) -> bool:
+		"""Returns whether the action has finished or not"""
+		...
+		
+	
+
```
alvarofpp: Ideally just one empty line at the end of each file.

alvarofpp (changes requested):

urnai/actions/chain_of_actions_base.py
CinquilCinquil: Should i remove those methods or implement them using the variable that will contain the actions?

alvarofpp: If the variable is a data structure that already has these methods, I believe you can remove the methods.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,37 @@
...
+		
+		if self.get_len() == 0: 
+			return None
+	
+		first_action = self.action_list[0]
+		
+		if first_action.is_complete():
+			self.action_list.pop(0)
+			self.next()
+		else:
```
alvarofpp: It can be removed.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,37 @@
...
+
+	"""This class works as a sequence of actions"""
+	
+	action_list: List[ActionBase]
+	
+	def __init__(self):
+		self.action_list = []
+	
+	def next(self):
+		# Executing the next action in the sequence
```
alvarofpp: The comment about a function must be made using `'''` or `"""`, otherwise Python will not recognize it in the `help` function.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,37 @@
...
+		first_action = self.action_list[0]
+		
+		if first_action.is_complete():
+			self.action_list.pop(0)
+			self.next()
+		else:
+			first_action.run()
+	
+	def check(self) -> bool:
+		# Checking all of the actions in the sequence
```
alvarofpp: The comment about a function must be made using `'''` or `"""`, otherwise Python will not recognize it in the `help` function.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,37 @@
...
+		
+		return all(action.check() for action in self.action_list)
+
+	def get_len(self) -> int:
+		return len(self.action_list)
```
alvarofpp: It could be an `@property`.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,37 @@
...
+		# Executing the next action in the sequence
+		
+		if self.get_len() == 0: 
+			return None
+	
+		first_action = self.action_list[0]
+		
+		if first_action.is_complete():
+			self.action_list.pop(0)
+			self.next()
```
alvarofpp: I don't agree with this approach. The way it is, I wouldn't be able to reuse an `ChainOfActionsBase` object, because the list of actions would be empty at the end. I think we could use an index to just select the actions, which would allow us to reuse the object in several executions.

alvarofpp (changes requested):
The unit tests are missing.

urnai/actions/chain_of_actions_base.py
```diff
@@ -14,24 +14,14 @@ class ChainOfActionsBase(ABC):
...
-	
-		first_action = self.action_list[0]
-		
-		if first_action.is_complete():
-			self.action_list.pop(0)
-			self.next()
-		else:
-			first_action.run()
+	def get_action(self, ind):
+		return self.action_list[ind]
```
alvarofpp:
```suggestion
	def get_action(self, action_index):
		return self.action_list[action_index]
```

urnai/actions/chain_of_actions_base.py
```diff
@@ -14,24 +14,14 @@ class ChainOfActionsBase(ABC):
...
 	
 	def check(self) -> bool:
-		# Checking all of the actions in the sequence
+		""" Checking all of the actions in the sequence """
 		
 		return all(action.check() for action in self.action_list)
 
+	@property
 	def get_len(self) -> int:
```
alvarofpp:
```suggestion
def len(self) -> int:
```

alvarofpp (changes requested):

CinquilCinquil:
I'm encountering this [error](https://github.com/UFRN-URNAI/urnai-tools/files/15129435/unit_test_error.txt) when i try to run `task unit-test`.

CinquilCinquil:
> I'm encountering this [error](https://github.com/UFRN-URNAI/urnai-tools/files/15129435/unit_test_error.txt) when i try to run `task unit-test`.

Its fixed, thanks to RickFqt 🥳 . The solution involves using `pytest` to run the tests.

tests/units/actions/test_action_base.py
```diff
@@ -0,0 +1,22 @@
...
+
+class TestActionBase(unittest.TestCase):
+
+    def test_abstract_methods(self):
+        ActionBase.__abstractmethods__ = set()
+
+        class FakeAction(ActionBase):
+            ...
+
+        f = FakeAction()
```
alvarofpp: The variable must be explicit.
```suggestion
        fake_action = FakeAction()
```

tests/units/actions/test_chain_of_actions_base.py
```diff
@@ -0,0 +1,23 @@
...
+class TestChainOfActionsBase(unittest.TestCase):
+
+	def test_abstract_methods(self):
+		ChainOfActionsBase.__abstractmethods__ = set()
+
+		class FakeChainOfActions(ChainOfActionsBase):
+			def __init__(self):
+				super().__init__()
+
+		f = FakeChainOfActions()
```
alvarofpp: The variable must be explicit.
```suggestion
		fake_chain_of_actions = FakeChainOfActions()
```

tests/units/actions/test_chain_of_actions_base.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class TestChainOfActionsBase(unittest.TestCase):
+
+	def test_abstract_methods(self):
+		ChainOfActionsBase.__abstractmethods__ = set()
+
+		class FakeChainOfActions(ChainOfActionsBase):
+			def __init__(self):
+				super().__init__()
```
alvarofpp: I think it would be better to place this class outside the `TestChainOfActionsBase`, so it's easier to reuse it for future tests if the need arises.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,28 @@
...
+	
+	action_list: list[ActionBase]
+	
+	def __init__(self):
+		self.action_list = []
+	
+	def get_action(self, action_index):
+		if action_index >= self.len:
+			return None
+		else:
```
alvarofpp: This `else` is not necessary.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,28 @@
...
+			return self.action_list[action_index]
+	
+	def check(self) -> bool:
+		""" Checking all of the actions in the sequence """
+		
+		return all(action.check() for action in self.action_list)
+
+	@property
+	def len(self) -> int:
```
alvarofpp: It can be `length`.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC
+
+from urnai.actions.action_base import ActionBase
+
+
+class ChainOfActionsBase(ABC):
+
+	"""This class works as a sequence of actions"""
+	
+	action_list: list[ActionBase]
```
alvarofpp, line 10: Does this declaration work? I couldn't identify where the `list` is declared.

urnai/actions/action_base.py
```diff
@@ -0,0 +1,20 @@
+from abc import ABC, abstractmethod
+
+
+class ActionBase(ABC):
+	id = None
+	
+	@abstractmethod
+	def run(self):
```
alvarofpp:
```suggestion
	def run(self) -> None:
```

urnai/actions/action_base.py
```diff
@@ -0,0 +1,20 @@
+from abc import ABC, abstractmethod
+
+
+class ActionBase(ABC):
+	id = None
```
alvarofpp: An `__init__` method to assign the value of `id` would be nice.

alvarofpp (changes requested):

tests/units/actions/test_chain_of_actions_base.py
CinquilCinquil: Does this apply to other tests similar to this one, such as the test for ActionBase?

urnai/actions/chain_of_actions_base.py
CinquilCinquil: It works for python 3.9 or greater. Its the actual `list` type, i found it [here](https://stackoverflow.com/questions/64496437/python-list-type-declaration).

urnai/actions/action_base.py
CinquilCinquil: I'm confused, you mentioned something else in a previous comment. Its the first change requested on March 9th, i set it to unsolved.

urnai/actions/action_base.py
```diff
@@ -0,0 +1,20 @@
...
+	
+	@abstractmethod
+	def check(self, obs) -> bool:
+		"""Returns whether the action can be executed or not"""
+		...
+	
+	@abstractmethod
+	def is_complete(self) -> bool:
+		"""Returns whether the action has finished or not"""
+		...
```
alvarofpp, line 21: Can be `@property`.

urnai/actions/chain_of_actions_base.py
```diff
@@ -0,0 +1,26 @@
...
+		self.action_list = []
+	
+	def get_action(self, action_index) -> ActionBase:
+		if action_index < self.length:
+			self.action_list[action_index]
+	
+	def check(self) -> bool:
+		""" Checking all of the actions in the sequence """
+		
+		return all(action.check() for action in self.action_list)
```
alvarofpp: Something that crossed my mind about this method is that there may be cases where an action can't be executed in that iteration, but would be available in the next one. Using `all` would then return `False`, because all `check()` must be `True` for it to return `True`.

It would be interesting to discuss this function at the next meeting to see if this logic is appropriate or not.

alvarofpp (changes requested):

alvarofpp (approved):

PR #80, opened 2024-03-13
Resolves #75

### LINKED ISSUE #75: Create base class for models
We need to have a base class for the DRL models that will be used by the agents.

Suggestions:

- Class name: `ModelBase`.
- Class path: `urnai.models`.

CinquilCinquil:
#75

### DIFF
--- a/urnai/models/model_base.py
+++ b/urnai/models/model_base.py
@@ -1,20 +1,24 @@
 from abc import ABC, abstractmethod
+from typing import Dict
 
 from urnai.utils.returns import ActionIndex
 
 
 class ModelBase(ABC):
 
-	learning_data = {}
+	learning_data: Dict
+	
+	def __init__(self):
+		self.learning_data = {}
 	
 	@abstractmethod
-	def learn(self, s, a, r, s_):
+	def learn(self, current_state, action, reward, next_state):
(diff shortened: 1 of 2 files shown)


urnai/models/model_base.py
```diff
@@ -0,0 +1,28 @@
...
+
+from urnai.utils.returns import ActionIndex
+
+
+class ModelBase(ABC):
+
+	learning_data = {}
+	
+	@abstractmethod
+	def learn(self, s, a, r, s_):
```
alvarofpp:
```suggestion
	def learn(self, current_state, action, reward, next_state):
```

urnai/models/model_base.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC, abstractmethod
+
+from urnai.utils.returns import ActionIndex
+
+
+class ModelBase(ABC):
+
+	learning_data = {}
```
alvarofpp: This is one of Python's little tricks.

ModelBase-test

Suggestion:

ModelBase-suggestion

alvarofpp (changes requested):
Unit tests are missing.

alvarofpp (approved):

PR #82, opened 2024-03-19
Resolves #73

### LINKED ISSUE #73: Create state representation class for StarCraft 2 races
From the class created in issue #72, it is necessary to create the state classes for the Protoss, Terran, and Zerg races.

Suggestions:

- Class names: `ProtossState`, `TerranState`, and `ZergState`.
- Class paths: `urnai.sc2.states`.
- Source: https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/states/sc2.py

RickFqt:
Resolves #73. When complete, this pull request should have:

- [x] Implementation of `ProtossState`, `TerranState`, and `ZergState` classes.
- [x] Implementation of unit tests for these classes.


RickFqt:
@alvarofpp For the implementation of `TerranState` class, I used `Simple64GridState` class from the [original sc2 state classes from urnai](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/states/sc2.py#L288) as a base. It basically stores information from the player (number of units, barracks and minerals, for example), and also builds a 4x4 grid that stores the location of enemy units. Do you think there is a better choice? I considered choosing [UnitStackingState](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/states/sc2.py#L495) as a base since it is pretty well commented, but it seemed to me that it is too specific for a general purpose class.

Also, I implemented the methods `get_my_units_amount` and `get_units_by_type` at the end of the file. Is that right?

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+        self._state = None
+
+    def update(self, obs):
+        if obs.game_loop[0] < 80 and self.base_top_left is None:
+
+            commandcenter = get_units_by_type(obs, units.Terran.CommandCenter)
+
+            if len(commandcenter) > 0:
+                townhall = commandcenter[0]
+                self.player_race = sc2_env.Race.terran
```
alvarofpp: Why is that if you already have the same declaration in `__init__`?

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+        new_state.extend(enemy_grid.flatten())
+        new_state.extend(player_grid.flatten())
+        final_state = np.expand_dims(new_state, axis=0)
+
+        self._state = final_state
+        return final_state
+
+    @property
+    def state(self):
+        return self._state
```
alvarofpp: Why not `self.state`?

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @property
+    def dimension(self):
+        return self._state_size
```
alvarofpp: Why not `self.dimension` or `self.size`?

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, grid_size=4):
+
+        self.grid_size = grid_size
+        # Size of the state returned
+        # 22: number of data added to the state (number of minerals, army_count, etc)
+        # 2 * 4 ** 2: 2 grids of size 4 x 4, representing enemy and player units
+        self._state_size = int(22 + 2 * (self.grid_size ** 2))
+        self.player_race = sc2_env.Race.terran
+        self.base_top_left = None
+        self._state = None
```
alvarofpp: You can put `self.reset()` to avoid having duplicate code.

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+from pysc2.env import sc2_env
+from pysc2.lib import features, units
+
+from urnai.states.state_base import StateBase
+
+
+class TerranState(StateBase):
+    
+    def __init__(self, grid_size=4):
+
```
alvarofpp, line 10: Remove blank line.

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+        # 22: number of data added to the state (number of minerals, army_count, etc)
+        # 2 * 4 ** 2: 2 grids of size 4 x 4, representing enemy and player units
+        self._state_size = int(22 + 2 * (self.grid_size ** 2))
+        self.player_race = sc2_env.Race.terran
+        self.base_top_left = None
+        self._state = None
+
+    def update(self, obs):
+        if obs.game_loop[0] < 80 and self.base_top_left is None:
+
```
alvarofpp: Remove blank line.

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+        enemy_grid = np.zeros((self.grid_size, self.grid_size))
+        player_grid = np.zeros((self.grid_size, self.grid_size))
+
+        enemy_units = [unit for unit in obs.raw_units if
+                       unit.alliance == features.PlayerRelative.ENEMY]
+        player_units = [unit for unit in obs.raw_units if
+                        unit.alliance == features.PlayerRelative.SELF]
+
+        for i in range(0, len(enemy_units)):
```
alvarofpp: `i` should be `enemy_index`.

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+        # Adds general information from the player.
+        new_state.append(obs.player.minerals / 6000)
+        new_state.append(obs.player.vespene / 6000)
+        new_state.append(obs.player.food_cap / 200)
+        new_state.append(obs.player.food_used / 200)
+        new_state.append(obs.player.food_army / 200)
+        new_state.append(obs.player.food_workers / 200)
+        new_state.append((obs.player.food_cap - obs.player.food_used) / 200)
+        new_state.append(obs.player.army_count / 200)
+        new_state.append(obs.player.idle_worker_count / 200)
```
alvarofpp: Why not something like:
```
new_state = [
  obs.player.minerals / 6000,
  obs.player.vespene / 6000,
  obs.player.food_cap / 200,
  ...
]
```
?

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+    
+    def __init__(self, grid_size=4):
+
+        self.grid_size = grid_size
+        # Size of the state returned
+        # 22: number of data added to the state (number of minerals, army_count, etc)
+        # 2 * 4 ** 2: 2 grids of size 4 x 4, representing enemy and player units
+        self._state_size = int(22 + 2 * (self.grid_size ** 2))
+        self.player_race = sc2_env.Race.terran
+        self.base_top_left = None
```
alvarofpp: Why `base_top_left`?

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+        new_state.append(get_my_units_amount(obs, units.Terran.FusionCore))
+        new_state.append(get_my_units_amount(obs, units.Terran.GhostAcademy))
+        new_state.append(get_my_units_amount(obs, units.Terran.Barracks) / 3)
+        new_state.append(get_my_units_amount(obs, units.Terran.Factory) / 2)
+        new_state.append(get_my_units_amount(obs, units.Terran.Starport) / 2)
+
+        # Instead of making a vector for all coordnates on the map, we'll
+        # discretize our enemy space
+        # and use a 4x4 grid to store enemy positions by marking a square as 1 if
+        # there's any enemy on it.
```
alvarofpp: The following piece of code could be in a separate function and only invoked here.

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,121 @@
...
+        self.base_top_left = None
+
+def get_my_units_amount(obs, unit_type):
+    return len(get_units_by_type(obs, unit_type, features.PlayerRelative.SELF))
+
+def get_units_by_type(obs, unit_type, alliance=features.PlayerRelative.SELF):
+    return [unit for unit in obs.raw_units
+            if unit.unit_type == unit_type
+            and unit.alliance == alliance
+            and unit.build_progress == 100]
```
alvarofpp: These functions should be kept in another file so that they can be reused by other classes.

alvarofpp (changes requested):

urnai/sc2/states/terran_state.py
RickFqt: I understood that this variable is a boolean that indicates if the townhall is located at the top left of the map. That would mean having an x and y with at most half of the size of the map, even though it only considers the x value being half of the map (as shown in the following):

```python
self.base_top_left = (townhall.x < 32)
```

I'm thinking of removing this variable, since I don't know what would be the benefit of keeping the townhall in the top left corner of the grid. Is that okay?

urnai/sc2/states/terran_state.py
RickFqt: About the following piece of code, I figured some possible mistakes. For example here:

```python
for i in range(0, len(enemy_units)):
            y = int(math.ceil((enemy_units[i].x + 1) / 64 / self.grid_size))
            x = int(math.ceil((enemy_units[i].y + 1) / 64 / self.grid_size))
            enemy_grid[x - 1][y - 1] += 1
```

I think the number `64` is for the current size of the map. So ideally, it shouldn't be a constant number, but an attribute from "obs" that represents the size of the map. 
Also, the parenthesis are incorrect, it should be like this:

```python
for i in range(0, len(enemy_units)):
            y = int(math.ceil((enemy_units[i].x + 1) / (64 / self.grid_size)))
            x = int(math.ceil((enemy_units[i].y + 1) / (64 / self.grid_size)))
            enemy_grid[x - 1][y - 1] += 1
```

urnai/sc2/states/terran_state.py
RickFqt: Should this file be inside the same folder as the `TerranState` class ?

urnai/sc2/states/terran_state.py
alvarofpp: I think we can remove this variable, as it restricts you to one map. We should leave this class as generic as possible, because other classes may arise from it that may need to know the position of townhall.

urnai/sc2/states/terran_state.py
alvarofpp: If 64 really does concern the size of the map, then we should remove it to make the class as generic as possible.

urnai/sc2/states/terran_state.py
alvarofpp: For now, I think so. If they are used by other classes that are not on the same path, then we can look at another place to put them.

urnai/sc2/states/terran_state.py
RickFqt: Is it line 11 or 13?

alvarofpp: 13

alvarofpp:
@RickFqt Please mark as solved the modifications that you have already solved.

urnai/sc2/states/terran_state.py
RickFqt: In #83 , the observation returned from our Environment Class has an attribute called "map_size", which would give what is needed to replace this `64` number. However, it would be better to discover the map's size from what is already returned by the pysc2 pure observation, right?

urnai/sc2/states/states_utils.py
```diff
@@ -0,0 +1,11 @@
+from pysc2.lib import features
+
+
+def get_my_units_amount(obs, unit_type):
+    return len(get_units_by_type(obs, unit_type, features.PlayerRelative.SELF))
+
+def get_units_by_type(obs, unit_type, alliance=features.PlayerRelative.SELF):
+    return [unit for unit in obs.raw_units
```
RickFqt: I was reviewing the #83 issue, and found that `obs` might not have the `raw_units` attribute (that happens when the `use_raw_units` parameter is set to `False` on the constructor). With that said, this current code could result in an error. I was thinking of changing it to something like this to prevent the error:
```python
def get_units_by_type(obs, unit_type, alliance=features.PlayerRelative.SELF):
    if has_attribute(obs, "raw_units"):
            return [unit for unit in obs.raw_units
                if unit.unit_type == unit_type
                and unit.alliance == alliance
                and unit.build_progress == 100]
    return []
```

Another solution could be to only call this method when `obs` has this attribute.

alvarofpp: Both approaches are good, so pick one.

urnai/sc2/states/terran_state.py
alvarofpp: > However, it would be better to discover the map's size from what is already returned by the pysc2 pure observation, right?

Yes, and I believe it's the ideal way forward if possible.

RickFqt: > Yes, and I believe it's the ideal way forward if possible.

Unfortunately, I don't think that's possible. Currently, this number corresponds to the `raw_resolution` of the map. In that case, only the `environment` class form pysc2 has access to it, and it doesn't add it as an attribute to a default `observation`. That said, the way I figured to solve this was to receive this `raw_resolution` as a parameter in the constructor of the `State` class.

urnai/sc2/states/states_utils.py
RickFqt: I chose the second approach. Now, the constructor receives a boolean called `use_raw_units`, just like on our sc2 `environment` class. Then, these methods are only called if this boolean is set to `True`.

urnai/sc2/states/terran_state.py
RickFqt: About the unit tests, should a different test file be created for these methods?

alvarofpp: Ideally, the unit test should be in a path similar to the file it is testing, so yes.

urnai/sc2/states/states_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Normalizing the values to always be between 0 and 1
+    # (since the max amount of units in SC2 is 200)
+    enemy_grid = enemy_grid / 200
+    player_grid = player_grid / 200
+
+    new_state.extend(enemy_grid.flatten())
+    new_state.extend(player_grid.flatten())
+
+    return new_state
```
alvarofpp: Lots of repeated code for players and enemies. I think there could be a function that receives a list of units and returns the grid.

urnai/sc2/states/states_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+def get_my_raw_units_amount(obs, unit_type):
+    return len(get_raw_units_by_type(obs, unit_type, features.PlayerRelative.SELF))
+
+def get_raw_units_by_type(obs, unit_type, alliance=features.PlayerRelative.SELF):
+    return [unit for unit in obs.raw_units
+            if unit.unit_type == unit_type
+            and unit.alliance == alliance
+            and unit.build_progress == 100]
```
alvarofpp: It's good practice not to have "magic numbers". I recommend creating a constant to represent this `100`.

urnai/sc2/states/states_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    for player_index in range(0, len(player_units)):
+        y = int(math.ceil((player_units[player_index].x + 1) / raw_to_grid_ratio))
+        x = int(math.ceil((player_units[player_index].y + 1) / raw_to_grid_ratio))
+        player_grid[x - 1][y - 1] += 1
+
+    # Normalizing the values to always be between 0 and 1
+    # (since the max amount of units in SC2 is 200)
+    enemy_grid = enemy_grid / 200
+    player_grid = player_grid / 200
```
alvarofpp: It's good practice not to have "magic numbers". I recommend creating a constant to represent this `200`.
Example:
```suggestion
    enemy_grid = enemy_grid / MAX_UNITS
    player_grid = player_grid / MAX_UNITS
```

alvarofpp (changes requested):
Many parts of the code have "magic numbers" (values without an explicit definition). I listed a few in the review, but I think you should create a file with constants to map them all.

RickFqt:
> Many parts of the code have "magic numbers" (values without an explicit definition). I listed a few in the review, but I think you should create a file with constants to map them all.

Should this new file be on the path `sc2/states/` as well?

alvarofpp:
> Should this new file be on the path sc2/states/ as well?

I don't think so. Constants are generally used in many parts of the project, so it is recommended to apply them to a single file or use `Enum` to give them more context.

RickFqt:
> I don't think so. Constants are generally used in many parts of the project, so it is recommended to apply them to a single file or use `Enum` to give them more context.

In that case, is the path `urnai/constants.py` a good choice? On the previous version, it was located on `urnai/utils/constants.py`, but I think we are not using `utils` anymore, right?

alvarofpp:
> In that case, is the path urnai/constants.py a good choice?

Yes.

urnai/sc2/states/states_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    return new_state
+
+def append_grid(
+        obs : list,
+        new_state : list,
+        grid_size : int, 
+        raw_resolution : int,
+        alliance : features.PlayerRelative,
+    ) -> list:
```
alvarofpp:
```suggestion
def append_alliance_grid(
        alliance : features.PlayerRelative,
        obs : list,
        new_state : list,
        grid_size : int, 
        raw_resolution : int,
    ) -> list:
```

urnai/sc2/states/states_utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Normalizing the values to always be between 0 and 1
+    grid = grid / SC2Constants.MAX_UNITS
+
+    new_state.extend(grid.flatten())
+
+    return new_state
+
+def get_my_raw_units_amount(obs : list, unit_type : int) -> int:
+    return len(get_raw_units_by_type(obs, unit_type, features.PlayerRelative.SELF))
```
alvarofpp:
```suggestion
def get_raw_units_amount(obs : list, unit_type : int, alliance : features.PlayerRelative = features.PlayerRelative.SELF,) -> int:
    return len(get_raw_units_by_type(obs, unit_type, alliance))
```

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,81 @@
...
+                    get_my_raw_units_amount(obs, units.Terran.EngineeringBay),
+                    get_my_raw_units_amount(obs, units.Terran.Armory),
+                    get_my_raw_units_amount(obs, units.Terran.MissileTurret) / 4,
+                    get_my_raw_units_amount(obs, units.Terran.SensorTower) / 1,
+                    get_my_raw_units_amount(obs, units.Terran.Bunker) / 4,
+                    get_my_raw_units_amount(obs, units.Terran.FusionCore),
+                    get_my_raw_units_amount(obs, units.Terran.GhostAcademy),
+                    get_my_raw_units_amount(obs, units.Terran.Barracks) / 3,
+                    get_my_raw_units_amount(obs, units.Terran.Factory) / 2,
+                    get_my_raw_units_amount(obs, units.Terran.Starport) / 2,
```
alvarofpp: This function goes through the same list several times just to perform a unit filter. Wouldn't it be better to turn this list into a dictionary using the unit type as the key? You can use `defaultdict` to return `0` in cases where there are no units of a given type.

urnai/sc2/states/states_utils.py
alvarofpp, line 1: `states_utils` is redundant because it is inside the `urnai/sc2/states` directory.

```
urnai/sc2/states/utils.py
```

tests/units/sc2/states/test_aux_methods_sc2_state.py
alvarofpp, line 1: It is interesting that the test files have the same name as the file you are testing, plus the prefix `test_`. This way we can easily find out where the tests for a particular file are.

tests/units/sc2/states/test_sc2_state.py
alvarofpp, line 1: This file contains tests for several different files. Each file must have its own test file.

alvarofpp (changes requested):

urnai/sc2/states/terran_state.py
RickFqt: Do you mean creating a dictionary that maps the types present in `obs.raw_units` to the number of times it appears?

RickFqt: This is how I tried it:

```python
def default_value():
    return 0

def create_raw_units_amount_dict(
        obs : list,
        alliance : features.PlayerRelative = features.PlayerRelative.SELF
    ) -> defaultdict:
    dict = defaultdict(default_value)
    for unit in obs.raw_units:
        if (unit.alliance == alliance
        and unit.build_progress == SC2Constants.BUILD_PROGRESS_COMPLETE):
            dict[unit.unit_type] += 1
    return dict
```

alvarofpp (changes requested):
The `ProstossState`, `TerranState` and `ZergState` classes have a lot of repeated code. I think it would be better to create a `StarCraft2State` class with this repeated code and have the race classes inherit from it.

urnai/sc2/states/terran_state.py
```diff
@@ -0,0 +1,49 @@
...
+                raw_units_amount_dict[units.Terran.EngineeringBay],
+                raw_units_amount_dict[units.Terran.Armory],
+                raw_units_amount_dict[units.Terran.MissileTurret] / 4,
+                raw_units_amount_dict[units.Terran.SensorTower] / 1,
+                raw_units_amount_dict[units.Terran.Bunker] / 4,
+                raw_units_amount_dict[units.Terran.FusionCore],
+                raw_units_amount_dict[units.Terran.GhostAcademy],
+                raw_units_amount_dict[units.Terran.Barracks] / 3,
+                raw_units_amount_dict[units.Terran.Factory] / 2,
+                raw_units_amount_dict[units.Terran.Starport] / 2,
```
alvarofpp: Magic numbers. Why aren't the values divided into Protoss and Zerg as they are in Terran?

alvarofpp (changes requested):

urnai/sc2/states/terran_state.py
RickFqt: According to Charles, these numbers are due to some experiments made by previous developers on urnai. He told me to leave them as it was at the time, so afterwards we could add some file (like the constants one) that imports these numbers. Do you think they should be removed?

alvarofpp: I would remove these numbers to make it like the class for Protoss and Zerg. You could create another class that overrides the `update` method of TerranState or develop a mechanism that allows you to change only the value of `units_amount_info`. I prefer the second option, but check the feasibility of both.

RickFqt: For now, I'll just remove these numbers and create another issue to solve this. Is that ok?

alvarofpp: Ok 👍🏻

alvarofpp (approved):

PR #84, opened 2024-04-05
Resolves #81

### LINKED ISSUE #81: Create a base class for saving and loading

CinquilCinquil:
#81

### DIFF
--- a/urnai/base/persistence.py
+++ b/urnai/base/persistence.py
@@ -30,33 +30,26 @@ class Persistence(ABC):
         if self.threaded_saving:
             self._threaded_save(savepath)
         else:
-            self._save(savepath)
+            self._simple_save(savepath)
 
     def _threaded_save(self, savepath):
         """
         This method saves pickleable
-        stuff in a separate thread
+        elements in a separate thread
         """
-        new_process = Process(target=self._save, args=(savepath,))
+        new_process = Process(target=self._simple_save, args=(savepath,))
         new_process.start()
 
     @abstractmethod
-    def _save(self, savepath):
+    def _simple_save(self, savepath):
         """
         This method handles logic related
-        to the child class
+        to non-threaded saving in the child class
         """
         ...
 
-    def load(self, loadpath):
-        self._load(loadpath)
-
     @abstractmethod
-    def _load(self, savepath):
-        """
-        This method handles logic related
-        to the child class
-        """
+    def load(self, savepath):
         ...
 
     @abstractmethod
--- a/urnai/base/persistence.py
+++ b/urnai/base/persistence.py
@@ -1,10 +1,8 @@
 import os
-import pickle
-import tempfile
-from multiprocessing import Process
+from abc import ABC, abstractmethod
 
 
-class Persistence:
+class Persistence(ABC):
     """
     This interface represents the concept of a class that can be saved to disk.
     The heir class should define a constant or attribute as a default filename
@@ -23,146 +21,25 @@ class Persistence:
         """
         return self.__class__.__name__ + '_'
 
-    def _save_pickle(self, persist_path):
-        """
-        This method saves our instance
-        using pickle.
-
-        First it checks which attributes should be
-        saved using pickle, the ones which are not
-        are backuped.
-
-        Then all unpickleable attributes are set to None
-        and the object is pickled.
-
-        Finally the nulled attributes are
-        restored.
-        """
-        path = self.get_full_persistance_path(persist_path)
-		
-        os.makedirs(os.path.dirname(path), exist_ok=True)
-		
-        with open(path, 'wb') as pickle_out:
-            pickle.dump(self._get_pickleable_dict(), pickle_out)
-
-    def _save_extra(self, persist_path):
-        """
-        This method should be implemented when
-        some extra persistence is to be saved.
-        """
-        pass
-
-    def _load_pickle(self, persist_path):
-        """
-        This method loads a list instance
-        saved by pickle.
-        """
-        pickle_path = self.get_full_persistance_path(persist_path)
-        exists_pickle = os.path.isfile(pickle_path)
-        
-        if exists_pickle and os.path.getsize(pickle_path) > 0:
-            with open(pickle_path, 'rb') as pickle_in:
-                pickle_dict = pickle.load(pickle_in)
-                self._restore_attributes(pickle_dict)
-
-    def _load_extra(self, persist_path):
-        """
-        This method should be implemented when
-        some extra persistence is to be loaded.
-        """
-        pass
-
     def get_full_persistance_path(self, persist_path):
         """This method returns the default persistance path."""
         return persist_path + os.path.sep + self._get_default_save_stamp()
 
+    @abstractmethod
     def save(self, savepath):
         """
         This method saves pickle objects
         and extra stuff needed
         """
-        
-        if not self.threaded_saving:
-            self._save_pickle(savepath)
-        else:
-            self._threaded_pickle_save(savepath)
+        ...
 
-        self._save_extra(savepath)
-
-    def _threaded_pickle_save(self, savepath):
-        """
-        This method saves pickleable
-        stuff in a separate thread
-        """
-		
-        new_process = Process(target=self._save_pickle, args=(savepath,))
-        new_process.start()
-
+    @abstractmethod
     def load(self, loadpath):
         """
         This method loads pickle objects
         and extra stuff needed
         """
-        self._load_pickle(loadpath)
-        self._load_extra(loadpath)
-
-    def _get_pickleable_attributes(self):
-        """
-        This method returns a list of pickeable attributes.
-		If you wish to block one particular pickleable attribute, put it
-        in self.attr_block_list as a string.
-        """
-        if not hasattr(self, 'attr_block_list') or self.attr_block_list is None:
-            self.attr_block_list = []
-
-        full_attr_list = [attr for attr in dir(self) if not attr.startswith('__')
-                          and not callable(getattr(self, attr))
-                          and attr not in self.attr_block_list
-                          and attr != 'attr_block_list']
-        pickleable_list = []
-
-        for key in full_attr_list:
-            try:
-                with tempfile.NamedTemporaryFile() as tmp_file:
-                    pickle.dump(getattr(self, key), tmp_file)
-                    tmp_file.flush()
-                
-                pickleable_list.append(key)
-                
-            except pickle.PicklingError:
-                continue
-            
-            except TypeError as type_error:
-                if ("can't pickle" not in str(type_error) or
-                 'cannot pickle' not in str(type_error)):
-                    raise
-                continue
-            
-            except NotImplementedError as notimpl_error:
-                if (str(notimpl_error) != 
-                'numpy() is only available when eager execution is enabled.'):
-                    raise
-                continue
-            
-            except AttributeError as attr_error:
-                if ("Can't pickle" in str(attr_error) or 
-                "object has no attribute '__getstate__'"):
-                    continue
-            
-            except ValueError as value_error:
-                if 'ctypes objects' not in str(value_error):
-                    raise
-                continue
-
-        return pickleable_list
-
-    def _get_pickleable_dict(self):
-        pickleable_attr_dict = {}
-
-        for attr in self._get_pickleable_attributes():
-            pickleable_attr_dict[attr] = getattr(self, attr)
-
-        return pickleable_attr_dict
+        ...
 
     def _restore_attributes(self, dict_to_restore):
         for key in dict_to_restore:
--- a/urnai/base/persistence.py
+++ b/urnai/base/persistence.py
@@ -1,5 +1,6 @@
 import os
 from abc import ABC, abstractmethod
+from multiprocessing import Process
 
 
 class Persistence(ABC):
@@ -25,19 +26,54 @@ class Persistence(ABC):
         """This method returns the default persistance path."""
         return persist_path + os.path.sep + self._get_default_save_stamp()
 
-    @abstractmethod
     def save(self, savepath):
+        if self.threaded_saving:
+            self._threaded_save(savepath)
+        else:
+            self._save(savepath)
+
+    def _threaded_save(self, savepath):
         """
-        This method saves pickle objects
-        and extra stuff needed
+        This method saves pickleable
+        stuff in a separate thread
+        """
+        new_process = Process(target=self._save, args=(savepath,))
+        new_process.start()
+
+    @abstractmethod
+    def _save(self, savepath):
+        """
+        This method handles logic related
+        to the child class
+        """
+        ...
+
+    def load(self, loadpath):
+        self._load(loadpath)
+
+    @abstractmethod
+    def _load(self, savepath):
+        """
+        This method handles logic related
+        to the child class
         """
         ...
 
     @abstractmethod
-    def load(self, loadpath):
+    def _get_dict(self):
         """
-        This method loads pickle objects
-        and extra stuff needed
+        This method returns a dict with
+        all the attributes that will be
+        saved
+        """
+        ...
+
+    @abstractmethod
+    def _get_attributes(self):
+        """
+        This method returns all of the
+        attribute names that can be saved
+        except those in blocklist
         """
         ...
 
--- a/urnai/base/persistence.py
+++ b/urnai/base/persistence.py
@@ -4,7 +4,7 @@ import tempfile
 from multiprocessing import Process
 
 
-class Savable:
+class Persistence:
     """
     This interface represents the concept of a class that can be saved to disk.
     The heir class should define a constant or attribute as a default filename
@@ -133,7 +133,8 @@ class Savable:
                 continue
             
             except TypeError as type_error:
-                if (not "can't pickle" in str(type_error)) or (not 'cannot pickle' in str(type_error)):
+                if ("can't pickle" not in str(type_error) or
+                 'cannot pickle' not in str(type_error)):
                     raise
                 continue
             
@@ -149,7 +150,7 @@ class Savable:
                     continue
             
             except ValueError as value_error:
-                if not 'ctypes objects' in str(value_error):
+                if 'ctypes objects' not in str(value_error):
                     raise
                 continue
 
--- a/urnai/base/persistence.py
+++ b/urnai/base/persistence.py
@@ -74,7 +74,5 @@ class Persistence(ABC):
 
     def _restore_attributes(self, dict_to_restore):
         for key in dict_to_restore:
-            if (key not in self.attr_block_list 
-                and key not in ('attr_block_list', 'processes')):
-                                
+            if (key not in self.attr_block_list):   
                 setattr(self, key, dict_to_restore[key])
--- a/urnai/base/persistence.py
+++ b/urnai/base/persistence.py
@@ -74,5 +74,5 @@ class Persistence(ABC):
 
     def _restore_attributes(self, dict_to_restore):
         for key in dict_to_restore:
-            if (key not in self.attr_block_list):   
+            if key not in self.attr_block_list:   
                 setattr(self, key, dict_to_restore[key])
--- a/urnai/base/persistence.py
+++ b/urnai/base/persistence.py
@@ -13,6 +13,7 @@ class Persistence(ABC):
     def __init__(self, threaded_saving=False):
         self.threaded_saving = threaded_saving
         self.attr_block_list = []
+        self.processes = []
 
     def _get_default_save_stamp(self):
         """
@@ -38,6 +39,7 @@ class Persistence(ABC):
         elements in a separate thread
         """
         new_process = Process(target=self._simple_save, args=(savepath,))
+        self.processes.append(new_process)
         new_process.start()
 
     @abstractmethod
@@ -72,5 +74,7 @@ class Persistence(ABC):
 
     def _restore_attributes(self, dict_to_restore):
         for key in dict_to_restore:
-            if key not in self.attr_block_list and key != 'attr_block_list':
+            if (key not in self.attr_block_list 
+                and key not in ('attr_block_list', 'processes')):
+                                
                 setattr(self, key, dict_to_restore[key])
--- a/urnai/base/persistence_pickle.py
+++ b/urnai/base/persistence_pickle.py
@@ -30,7 +30,7 @@ class PersistencePickle(Persistence):
         Finally the nulled attributes are
         restored.
         """
-        path = super().get_full_persistance_path(persist_path)
+        path = self.get_full_persistance_path(persist_path)
 		
         os.makedirs(os.path.dirname(path), exist_ok=True)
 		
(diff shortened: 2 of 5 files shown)


CinquilCinquil:
I saw that there was already a [savable](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/base/savable.py) class on the previous version of URNAI, so i tested it, fixed some bugs and i think its ok to use now.

urnai/base/savable.py
```diff
@@ -0,0 +1,225 @@
+import os
+import pickle
+import tempfile
+from multiprocessing import Process
+
+#import time
+#from urnai.utils.reporter import Reporter as rp
```
alvarofpp: Remove this.

urnai/base/savable.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class SavableAttr:
+    def __init__(self, value):
+        self.value = value
```
alvarofpp: I think we can remove this and use a `Dict` instead.

urnai/base/savable.py
```diff
@@ -0,0 +1,225 @@
...
+class Savable:
+    """
+    This interface represents the concept of a class that can be saved to disk.
+    The heir class should define a constant or attribute as a default filename
+	to save on disk.
+    """
+
+    def __init__(self, threaded_saving=False):
+        self.threaded_saving = threaded_saving
+        self.pickle_black_list = []
```
alvarofpp: `blacklist` is a racist term. Use `blocklist` instead.

alvarofpp: `pickle` restricts the type of file that will be saved. I would use a more general term, as we intend to use several types of files.

urnai/base/savable.py
```diff
@@ -0,0 +1,225 @@
...
+		+ 'tensorflow_')
+
+    def get_full_persistance_path(self, persist_path):
+        """This method returns the default persistance path."""
+        return persist_path + os.path.sep + self.get_default_save_stamp()
+
+    def get_full_persistance_pytorch_path(self, persist_path):
+        """This method returns the default persistance pytorch path."""
+        return (persist_path + os.path.sep + self.get_default_save_stamp()
+		+ '.pt')
```
alvarofpp: I would remove all these functions, with the exception of `get_full_persistance_path`.

alvarofpp (changes requested):
I noticed that `load` methods are in the `Savable` class, I think they should be in another class or we could change the name of `Savable` to something that encompasses these functions.

urnai/base/savable.py
CinquilCinquil: I noticed that this class wasn't being used for anything, so i just removed it, ok?

CinquilCinquil:
For testing, i woud suggest first creating a class, that inherits from Savable, with some attributes. Then instantiate it, giving random values to the attributes, after that save it and load it to see if it still has the same values as before. I belive all methods of the class are used in this procedure, and that should reveal if any of them are not functional.

CinquilCinquil:
Should i make unit tests?

urnai/base/savable.py
alvarofpp: Yes.

alvarofpp:
> For testing, i woud suggest first creating a class, that inherits from Savable, with some attributes. Then instantiate it, giving random values to the attributes, after that save it and load it to see if it still has the same values as before. I belive all methods of the class are used in this procedure, and that should reveal if any of them are not functional.

The Savable class is not abstract, so you should use it directly in the tests. As these are unit tests, you should ideally create tests for each function separately.

alvarofpp:
> Should i make unit tests?

Yes, you should.

CinquilCinquil:
What about the name Persistence for the class?

urnai/base/savable.py
```diff
@@ -155,18 +132,18 @@ def get_pickleable_attributes(self):
...
-            self.pickle_black_list = []
+        if not hasattr(self, 'attr_block_list') or self.attr_block_list is None:
+            self.attr_block_list = []
 
         full_attr_list = [a for a in dir(self) if not a.startswith('__')
                           and not callable(getattr(self, a))
-                          and a not in self.pickle_black_list
-                          and a != 'pickle_black_list']
+                          and a not in self.attr_block_list
+                          and a != 'attr_block_list']
```
alvarofpp: Rename `a` to something more explicit.

urnai/base/savable.py
```diff
@@ -155,18 +132,18 @@ def get_pickleable_attributes(self):
         """
         This method returns a list of
         pickeable attributes. If you wish
-        to blacklist one particular
+        to blocklist one particular
         pickleable attribute, put it
-        in self.pickle_black_list as a
+        in self.attr_block_list as a
         string.
```
alvarofpp: You can add more characters to the lines.

urnai/base/savable.py
```diff
@@ -218,8 +195,8 @@ def get_pickleable_dict(self):
 
     def restore_pickleable_attributes(self, dict_to_restore):
```
alvarofpp: Rename the function.

urnai/base/savable.py
```diff
@@ -218,8 +195,8 @@ def get_pickleable_dict(self):
...
         for attr in dict_to_restore:
-            if hasattr(self, 'pickle_black_list'):
-                if attr not in self.pickle_black_list and attr != 'pickle_black_list':
+            if hasattr(self, 'attr_block_list'):
```
alvarofpp: `attr_block_list` is started in `__init__`, so I can't imagine a case where it isn't in `self`.

urnai/base/savable.py
```diff
@@ -218,8 +195,8 @@ def get_pickleable_dict(self):
 
     def restore_pickleable_attributes(self, dict_to_restore):
         for attr in dict_to_restore:
```
alvarofpp:
```suggestion
for key in dict_to_restore:
```

urnai/base/savable.py
```diff
@@ -218,8 +195,8 @@ def get_pickleable_dict(self):
...
-            if hasattr(self, 'pickle_black_list'):
-                if attr not in self.pickle_black_list and attr != 'pickle_black_list':
+            if hasattr(self, 'attr_block_list'):
+                if attr not in self.attr_block_list and attr != 'attr_block_list':
                     setattr(self, attr, dict_to_restore[attr])
             else:
                 setattr(self, attr, dict_to_restore[attr])
```
alvarofpp: Avoid `setattr`.

urnai/base/savable.py
```diff
@@ -218,8 +195,8 @@ def get_pickleable_dict(self):
 
     def restore_pickleable_attributes(self, dict_to_restore):
         for attr in dict_to_restore:
-            if hasattr(self, 'pickle_black_list'):
-                if attr not in self.pickle_black_list and attr != 'pickle_black_list':
+            if hasattr(self, 'attr_block_list'):
+                if attr not in self.attr_block_list and attr != 'attr_block_list':
                     setattr(self, attr, dict_to_restore[attr])
```
alvarofpp: Avoid `setattr`.

urnai/base/savable.py
```diff
@@ -155,18 +132,18 @@ def get_pickleable_attributes(self):
...
 
         full_attr_list = [a for a in dir(self) if not a.startswith('__')
                           and not callable(getattr(self, a))
-                          and a not in self.pickle_black_list
-                          and a != 'pickle_black_list']
+                          and a not in self.attr_block_list
+                          and a != 'attr_block_list']
         pickleable_list = []
 
         for attr in full_attr_list:
```
alvarofpp:
```suggestion
for key in full_attr_list:
```

alvarofpp (changes requested):

alvarofpp:
> What about the name Persistence for the class?

It's okay for now.

urnai/base/savable.py
CinquilCinquil: Any suggestions on how? i've found this [page](https://snyk.io/pt-BR/blog/the-dangers-of-setattr-avoiding-mass-assignment/) and it shows mutiple ways, can i choose any of them?

CinquilCinquil:
Btw, my lint checks keep failing because of the description of an old commit (I put too many characters in a line apparently). Is there any way to fix that?

urnai/base/savable.py
alvarofpp: You can use whichever approach you prefer. If there are only a few attributes, I recommend making it as explicit as possible. If it's a lot of attributes, I think it's better to validate the dictionary schema and then copy it.

alvarofpp:
> Btw, my lint checks keep failing because of the description of an old commit (I put too many characters in a line apparently). Is there any way to fix that?

You can use interactive rebase (`git rebase -i`) for that.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+    This interface represents the concept of a class that can be saved to disk.
+    The heir class should define a constant or attribute as a default filename
+	to save on disk.
+    """
+
+    def __init__(self, threaded_saving=False):
+        self.threaded_saving = threaded_saving
+        self.attr_block_list = []
+
+    def get_default_save_stamp(self):
```
alvarofpp: Most likely, this function will only be used by the class and not by the developer. I would add the prefix `_`.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+        verbosity_lvl=2)
+        '''
+
+    def threaded_pickle_save(self, savepath):
+        """
+        This method saves pickleable
+        stuff in a separate thread
+        """
+        #rp.report('THREADED Saving {} object...'
+		#.format(self.__class__.__name__), verbosity_lvl=1)
```
alvarofpp: Remove comment.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+            self.threaded_pickle_save(savepath)
+
+        self.save_extra(savepath)
+        
+        #end_time = time.time()
+        '''
+        rp.report('It took {} seconds to save {} object!'
+		.format(end_time - start_time, self.__class__.__name__),
+        verbosity_lvl=2)
+        '''
```
alvarofpp: Remove comment.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+        return persist_path + os.path.sep + self.get_default_save_stamp()
+
+    def save(self, savepath):
+        """
+        This method saves pickle objects
+        and extra stuff needed
+        """
+        #rp.report('Saving {} object...'.format
+		#(self.__class__.__name__), verbosity_lvl=1)
+        #start_time = time.time()
```
alvarofpp: Remove comment.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Then all unpickleable attributes are set to None
+        and the object is pickled.
+
+        Finally the nulled attributes are
+        restored.
+        """
+        path = self.get_full_persistance_path(persist_path)
+		
+		# creating necessary directories for the file
```
alvarofpp: Remove comment.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def get_default_save_stamp(self):
+        """
+        This method returns the default
+        file name that should be used while
+        persisting the object.
+        """
+        return self.__class__.__name__ + '_'
+
+    def save_pickle(self, persist_path):
```
alvarofpp: It should be a generalized method and then a class for the pickle format that would contain the code for that specific format.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+            pickle.dump(self.get_pickleable_dict(), pickle_out)
+
+    def save_extra(self, persist_path):
+        """
+        This method should be implemented when
+        some extra persistence is to be saved.
+        """
+        pass
+
+    def load_pickle(self, persist_path):
```
alvarofpp: It should be a generalized method and then a class for the pickle format that would contain the code for that specific format.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+            
+            except ValueError as ve:
+                if 'ctypes objects' in str(ve):
+                    continue
+                else:
+                    raise
+
+        return pickleable_list
+
+    def get_pickleable_dict(self):
```
alvarofpp: It should be a generalized method and then a class for the pickle format that would contain the code for that specific format.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def load(self, loadpath):
+        """
+        This method loads pickle objects
+        and extra stuff needed
+        """
+        self.load_pickle(loadpath)
+        self.load_extra(loadpath)
+
+    def get_pickleable_attributes(self):
```
alvarofpp: It should be a generalized method and then a class for the pickle format that would contain the code for that specific format.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+        self.save_extra(savepath)
+        
+        #end_time = time.time()
+        '''
+        rp.report('It took {} seconds to save {} object!'
+		.format(end_time - start_time, self.__class__.__name__),
+        verbosity_lvl=2)
+        '''
+
+    def threaded_pickle_save(self, savepath):
```
alvarofpp: It should be a generalized method and then a class for the pickle format that would contain the code for that specific format.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+        some extra persistence is to be saved.
+        """
+        pass
+
+    def load_pickle(self, persist_path):
+        """
+        This method loads a list instance
+        saved by pickle.
+        """
+        # Check if pickle file exists
```
alvarofpp: Remove comment.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+        pickle_path = self.get_full_persistance_path(persist_path)
+        exists_pickle = os.path.isfile(pickle_path)
+        
+        # If yes, load it
```
alvarofpp: Remove comment.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+            with open(pickle_path, 'rb') as pickle_in:
+                pickle_dict = pickle.load(pickle_in)
+                self.restore_attributes(pickle_dict)
+                '''
+                rp.report(
+                    '**************************************** \n Loaded pickle for '
+                    + self.get_default_save_stamp()
+                    + ' loaded. \n****************************************',
+                2)
+				'''
```
alvarofpp: Remove comment.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+
+        for attr in self.get_pickleable_attributes():
+            pickleable_attr_dict[attr] = getattr(self, attr)
+
+        return pickleable_attr_dict
+
+    def restore_attributes(self, dict_to_restore):
+        for key in dict_to_restore:
+            if key not in self.attr_block_list and key != 'attr_block_list':
+                setattr(self, key, dict_to_restore[key])
```
alvarofpp: Avoid using `setattr`.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+            except AttributeError as ae:
+                if ("Can't pickle" in str(ae) or 
+                "object has no attribute '__getstate__'"):
+                    continue
+            
+            except ValueError as ve:
+                if 'ctypes objects' in str(ve):
+                    continue
+                else:
+                    raise
```
alvarofpp: Remove `else`, change the logic if necessary.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+            
+            except NotImplementedError as nie:
+                if (str(nie) == 
+                'numpy() is only available when eager execution is enabled.'):
+                    continue
+                else:
+                    raise
```
alvarofpp: Remove `else`, change the logic if necessary.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+            except pickle.PicklingError:
+                continue
+            
+            except TypeError as te:
+                if "can't pickle" in str(te):
+                    continue
+                if 'cannot pickle' in str(te):
+                    continue
+                else:
+                    raise
```
alvarofpp: Remove `else`, change the logic if necessary.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+        and extra stuff needed
+        """
+        #rp.report('Saving {} object...'.format
+		#(self.__class__.__name__), verbosity_lvl=1)
+        #start_time = time.time()
+        
+        if not self.threaded_saving:
+            self.save_pickle(savepath)
+        else:
+            self.threaded_pickle_save(savepath)
```
alvarofpp: Remove `else`, change the logic if necessary.

urnai/base/savable.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def threaded_pickle_save(self, savepath):
+        """
+        This method saves pickleable
+        stuff in a separate thread
+        """
+        #rp.report('THREADED Saving {} object...'
+		#.format(self.__class__.__name__), verbosity_lvl=1)
+		
+        p = Process(target=self.save_pickle, args=(savepath,))
```
alvarofpp: Change the variable name to something more explicit.

alvarofpp (changes requested):

urnai/base/savable.py
CinquilCinquil: Why?

alvarofpp: This was explained at the last meeting.

urnai/base/persistence.py
```diff
@@ -0,0 +1,83 @@
...
+    def load(self, loadpath):
+        self._load(loadpath)
+
+    @abstractmethod
+    def _load(self, savepath):
+        """
+        This method handles logic related
+        to the child class
+        """
+        ...
```
alvarofpp: Instead, you can simply set `load` to `abstractmethod`.

```suggestion
    @abstractmethod
    def load(self, loadpath):
        """
        This method handles logic related
        to the child class
        """
        ...
```

urnai/base/persistence.py
```diff
@@ -0,0 +1,83 @@
...
+        new_process = Process(target=self._save, args=(savepath,))
+        new_process.start()
+
+    @abstractmethod
+    def _save(self, savepath):
+        """
+        This method handles logic related
+        to the child class
+        """
+        ...
```
alvarofpp: To avoid confusion with the `save` method, I would rename this to something like `simple_save`.

alvarofpp (changes requested):

alvarofpp:
Unit tests are missing.

alvarofpp (changes requested):
The tests for `PersistencePickle` are missing.

CinquilCinquil:
In the `_simple_save` and `load` tests a file is created, but i wish to delete it after the test is done. I tried `os.remove` but it needs extra permission to work. Any suggestions?

alvarofpp:
@CinquilCinquil I recommend mocking the functions that create and read the file, that way the file won't be created and won't need to be removed later.

CinquilCinquil:
Oh, but the whole purpose of `test_simple_save`, for example, is to check if a file has been created. I'm not familiar with mocking, is  there a way of faking file creation?

alvarofpp:
Creating files is something that can result storage costs, and you have to make sure that each test involving `_simple_save` generates a different file so that you don't have the problem of trying to save to the same file or the file not existing because of another test. In short, it's something that can cause more problems than it solves. To test this, I believe that flow/process tests are more suitable than unit tests.

What you can do is mock the function that saves or reads the file and change its execution, in which case you should apply the mock to the `return_value` of the function.

urnai/base/persistence_pickle.py
```diff
@@ -0,0 +1,111 @@
...
+        saved using pickle, the ones which are not
+        are backuped.
+
+        Then all unpickleable attributes are set to None
+        and the object is pickled.
+
+        Finally the nulled attributes are
+        restored.
+        """
+        path = super().get_full_persistance_path(persist_path)
```
alvarofpp:
```suggestion
        path = self.get_full_persistance_path(persist_path)
```

urnai/base/persistence_pickle.py
```diff
@@ -0,0 +1,111 @@
...
+		
+        with open(path, 'wb') as pickle_out:
+            pickle.dump(self._get_dict(), pickle_out)
+
+    def load(self, persist_path):
+        """
+        This method loads a list instance
+        saved by pickle.
+        """
+        pickle_path = super().get_full_persistance_path(persist_path)
```
alvarofpp:
```suggestion
        pickle_path = self.get_full_persistance_path(persist_path)
```

urnai/base/persistence_pickle.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        if not hasattr(self, 'attr_block_list') or self.attr_block_list is None:
+            self.attr_block_list = []
+
+        full_attr_list = [attr for attr in dir(self) if not attr.startswith('__')
+                          and not callable(getattr(self, attr))
+                          and attr not in self.attr_block_list
+                          and attr != 'attr_block_list'
+                          and attr != 'processes'
+                          and 'abc' not in attr]
```
alvarofpp:
```suggestion
        if not hasattr(self, 'attr_block_list') or self.attr_block_list is None:
            self.attr_block_list = []

        attr_block_list = self.attr_block_list + ['attr_block_list', 'processes']

        full_attr_list = [attr for attr in dir(self) if not attr.startswith('__')
                          and not callable(getattr(self, attr))
                          and attr not in attr_block_list
                          and 'abc' not in attr]
```

urnai/base/persistence_pickle.py
```diff
@@ -0,0 +1,111 @@
...
+                continue
+            
+            except AttributeError as attr_error:
+                if ("Can't pickle" in str(attr_error) or 
+                "object has no attribute '__getstate__'"):
+                    continue
+            
+            except ValueError as value_error:
+                if 'ctypes objects' not in str(value_error):
+                    raise
```
alvarofpp: Raise what?

urnai/base/persistence_pickle.py
```diff
@@ -0,0 +1,111 @@
...
+                
+                pickleable_list.append(key)
+                
+            except pickle.PicklingError:
+                continue
+            
+            except TypeError as type_error:
+                if ("can't pickle" not in str(type_error) or
+                 'cannot pickle' not in str(type_error)):
+                    raise
```
alvarofpp: Raise what?

urnai/base/persistence_pickle.py
```diff
@@ -0,0 +1,111 @@
...
+                continue
+            
+            except NotImplementedError as notimpl_error:
+                if (str(notimpl_error) != 
+                'numpy() is only available when eager execution is enabled.'):
+                    raise
```
alvarofpp: Raise what?

alvarofpp (changes requested):

alvarofpp (approved):

PR #85, opened 2024-04-05
Resolves #83

### LINKED ISSUE #83: Create Environment class for Starcraft 2
Suggestions:

- Source: [https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/envs/sc2.py](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/envs/sc2.py)

RickFqt:
Resolves #83.

- [x] Creates an Environment class for Starcraft 2.
- [x] Implements unit tests for it.

### DIFF
--- a/tests/units/sc2/environments/test_sc2_environment.py
+++ b/tests/units/sc2/environments/test_sc2_environment.py
@@ -42,15 +42,22 @@ class TestSC2Environment(unittest.TestCase):
     @patch('pysc2.env.sc2_env.SC2Env')
     def test_reset_method(self, scenv_mock):
         # GIVEN
-        env = SC2Env(step_mul=16, game_steps_per_episode=100)
+        env = SC2Env(map_name='Simple64', step_mul=16, game_steps_per_episode=100)
+        fakeTimestep = ( TimeStep( 
+            step_type= StepType.FIRST, 
+            reward=0, 
+            discount=0, 
+            observation=NamedDict({})), 
+        )
         env.env_instance._episode_steps = 16
         env.env_instance._episode_length = 100
+        env.env_instance._interface_formats[0]._raw_resolution = point.Point(64,64)
+        env.env_instance.reset.return_value = fakeTimestep
         # WHEN
         obs = env.reset()
         # THEN
         env.env_instance.reset.assert_called_once()
-        isinstance(obs, NamedDict)
-        pass
+        assert isinstance(obs, NamedDict)
     
     # @patch('pysc2.env.sc2_env.SC2Env')
     # def test_step_method(self, scenv_mock):
@@ -98,9 +105,9 @@ class TestSC2Environment(unittest.TestCase):
         # WHEN
         obs, reward, terminated, truncated = env._parse_timestep(fakeTimestep)
         # THEN
-        isinstance(obs, NamedDict)
+        assert isinstance(obs, NamedDict)
         assert obs.step_mul == 16
-        isinstance(obs.map_size, point.Point)
+        assert isinstance(obs.map_size, point.Point)
         assert obs.map_size.x == 64
         assert obs.map_size.y == 64
         assert reward == 0
--- a/tests/units/sc2/environments/test_sc2_environment.py
+++ b/tests/units/sc2/environments/test_sc2_environment.py
@@ -1,6 +1,10 @@
 import unittest
 from unittest.mock import patch
 
+from pysc2.env.environment import StepType, TimeStep
+from pysc2.lib import point
+from pysc2.lib.named_array import NamedDict
+
 from urnai.sc2.environments.sc2environment import SC2Env
 
 
@@ -9,7 +13,10 @@ class TestSC2Environment(unittest.TestCase):
     @patch('pysc2.lib.features.AgentInterfaceFormat')
     @patch('pysc2.env.sc2_env.SC2Env')
     def test_start_on_init(self, MockSC2Env, MockAgentInterfaceFormat, MockDimensions):
+        # GIVEN any state
+        # WHEN
         SC2Env()
+        # THEN
         MockSC2Env.assert_called_once()
         MockAgentInterfaceFormat.assert_called_once()
         MockDimensions.assert_called_once()
@@ -22,32 +29,80 @@ class TestSC2Environment(unittest.TestCase):
     ):
         # Even after another call of start, the sc2_env of pysc2
         # should only be instanciated once.
+
+        # GIVEN
         env = SC2Env()
+        # WHEN
         env.start()
+        # THEN
         MockSC2Env.assert_called_once()
         MockAgentInterfaceFormat.assert_called_once()
         MockDimensions.assert_called_once()
 
-    def test_step_method(self):
-        # TODO
+    @patch('pysc2.env.sc2_env.SC2Env')
+    def test_reset_method(self, scenv_mock):
+        # GIVEN
+        env = SC2Env(step_mul=16, game_steps_per_episode=100)
+        env.env_instance._episode_steps = 16
+        env.env_instance._episode_length = 100
+        # WHEN
+        obs = env.reset()
+        # THEN
+        env.env_instance.reset.assert_called_once()
+        isinstance(obs, NamedDict)
         pass
+    
+    # @patch('pysc2.env.sc2_env.SC2Env')
+    # def test_step_method(self, scenv_mock):
+    #     # TODO
+    #     # GIVEN
+    #     env = SC2Env()
+    #     env.reset()
+    #     # WHEN
+    #     obs, reward, terminated, truncated = env.step([actions.RAW_FUNCTIONS.no_op()])
+    #     # THEN
+    #     env.env_instance.step.assert_called_once()
+    #     pass
 
-    def test_reset_method(self):
-        # TODO
-        pass
 
     @patch('pysc2.env.sc2_env.SC2Env')
-    def test_close_method(self, MockSC2Env):
-        # TODO: Check why this does not work
-        # env = SC2Env()
-        # env.close()
-        # MockSC2Env.close.assert_called_once()
-        pass
+    def test_close_method(self, scenv_mock):
+        # GIVEN
+        env = SC2Env()
+        # WHEN
+        env.close()
+        # THEN
+        env.env_instance.close.assert_called_once()
     
     def test_restart_method(self):
         # TODO
+        # GIVEN
+        # WHEN
+        # THEN
         pass
 
-    def test_parse_timestep_method(self):
-        # TODO
-        pass
+    @patch('pysc2.env.sc2_env.SC2Env')
+    def test_parse_timestep_method(self, scenv_mock):
+
+        # GIVEN
+        env = SC2Env(map_name='Simple64', step_mul=16, game_steps_per_episode=100)
+        fakeTimestep = ( TimeStep( 
+            step_type= StepType.FIRST, 
+            reward=0, 
+            discount=0, 
+            observation=NamedDict({})), 
+        )
+        env.env_instance._episode_steps = 16
+        env.env_instance._episode_length = 100
+        env.env_instance._interface_formats[0]._raw_resolution = point.Point(64,64)
+        # WHEN
+        obs, reward, terminated, truncated = env._parse_timestep(fakeTimestep)
+        # THEN
+        isinstance(obs, NamedDict)
+        assert obs.step_mul == 16
+        isinstance(obs.map_size, point.Point)
+        assert obs.map_size.x == 64
+        assert obs.map_size.y == 64
+        assert reward == 0
+        assert terminated is False
+        assert truncated is False
--- a/tests/units/sc2/environments/test_sc2_environment.py
+++ b/tests/units/sc2/environments/test_sc2_environment.py
@@ -1,8 +1,9 @@
 import unittest
 from unittest.mock import patch
 
+from pysc2.env import sc2_env
 from pysc2.env.environment import StepType, TimeStep
-from pysc2.lib import point
+from pysc2.lib import actions, point
 from pysc2.lib.named_array import NamedDict
 
 from urnai.sc2.environments.sc2environment import SC2Env
@@ -15,7 +16,10 @@ class TestSC2Environment(unittest.TestCase):
     def test_start_on_init(self, MockSC2Env, MockAgentInterfaceFormat, MockDimensions):
         # GIVEN any state
         # WHEN
-        SC2Env()
+        SC2Env(players=[
+                sc2_env.Agent(sc2_env.Race.protoss),
+                sc2_env.Bot(sc2_env.Race.terran, sc2_env.Difficulty.very_easy),
+            ])
         # THEN
         MockSC2Env.assert_called_once()
         MockAgentInterfaceFormat.assert_called_once()
@@ -31,7 +35,10 @@ class TestSC2Environment(unittest.TestCase):
         # should only be instanciated once.
 
         # GIVEN
-        env = SC2Env()
+        env = SC2Env(players=[
+                sc2_env.Agent(sc2_env.Race.terran),
+                sc2_env.Bot(sc2_env.Race.zerg, sc2_env.Difficulty.easy),
+            ])
         # WHEN
         env.start()
         # THEN
@@ -59,17 +66,32 @@ class TestSC2Environment(unittest.TestCase):
         env.env_instance.reset.assert_called_once()
         assert isinstance(obs, NamedDict)
     
-    # @patch('pysc2.env.sc2_env.SC2Env')
-    # def test_step_method(self, scenv_mock):
-    #     # TODO
-    #     # GIVEN
-    #     env = SC2Env()
-    #     env.reset()
-    #     # WHEN
-    #     obs, reward, terminated, truncated = env.step([actions.RAW_FUNCTIONS.no_op()])
-    #     # THEN
-    #     env.env_instance.step.assert_called_once()
-    #     pass
+    @patch('pysc2.env.sc2_env.SC2Env')
+    def test_step_method(self, scenv_mock):
+        # GIVEN
+        env = SC2Env(map_name='Simple64', step_mul=16, game_steps_per_episode=100)
+        fakeTimestep = ( TimeStep( 
+            step_type= StepType.FIRST, 
+            reward=0, 
+            discount=0, 
+            observation=NamedDict({})), 
+        )
+        env.env_instance._episode_steps = 16
+        env.env_instance._episode_length = 100
+        env.env_instance._interface_formats[0]._raw_resolution = point.Point(64,64)
+        env.env_instance.step.return_value = fakeTimestep
+        # WHEN
+        obs, reward, terminated, truncated = env.step([actions.RAW_FUNCTIONS.no_op()])
+        # THEN
+        env.env_instance.step.assert_called_once_with([actions.RAW_FUNCTIONS.no_op()])
+        assert isinstance(obs, NamedDict)
+        assert obs.step_mul == 16
+        assert isinstance(obs.map_size, point.Point)
+        assert obs.map_size.x == 64
+        assert obs.map_size.y == 64
+        assert reward == 0
+        assert terminated is False
+        assert truncated is False
 
 
     @patch('pysc2.env.sc2_env.SC2Env')
@@ -81,12 +103,25 @@ class TestSC2Environment(unittest.TestCase):
         # THEN
         env.env_instance.close.assert_called_once()
     
-    def test_restart_method(self):
-        # TODO
+    @patch('pysc2.env.sc2_env.SC2Env')
+    def test_restart_method(self, scenv_mock):
         # GIVEN
+        env = SC2Env(map_name='Simple64', step_mul=16, game_steps_per_episode=100)
+        fakeTimestep = ( TimeStep( 
+            step_type= StepType.FIRST, 
+            reward=0, 
+            discount=0, 
+            observation=NamedDict({})), 
+        )
+        env.env_instance._episode_steps = 16
+        env.env_instance._episode_length = 100
+        env.env_instance._interface_formats[0]._raw_resolution = point.Point(64,64)
+        env.env_instance.reset.return_value = fakeTimestep
         # WHEN
+        env.restart()
         # THEN
-        pass
+        env.env_instance.close.assert_called_once()
+        env.env_instance.reset.assert_called_once()
 
     @patch('pysc2.env.sc2_env.SC2Env')
     def test_parse_timestep_method(self, scenv_mock):
--- a/tests/units/sc2/environments/test_sc2_environment.py
+++ b/tests/units/sc2/environments/test_sc2_environment.py
@@ -1,27 +1,53 @@
 import unittest
-
-from pysc2.env import sc2_env
+from unittest.mock import patch
 
 from urnai.sc2.environments.sc2environment import SC2Env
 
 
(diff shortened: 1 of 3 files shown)


urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,134 @@
...
+from pysc2.env.environment import StepType
+from pysc2.lib import actions, features
+
+from urnai.environments.environment_base import EnvironmentBase
+
+sc2_races = {
+    'terran': sc2_env.Race.terran,
+    'protoss': sc2_env.Race.protoss,
+    'zerg': sc2_env.Race.zerg,
+    'random': sc2_env.Race.random}
```
alvarofpp: What is the difference between using `sc2_races` and `sc2_env.Race`?

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,134 @@
...
+    'very_easy': sc2_env.Difficulty.very_easy,
+    'easy': sc2_env.Difficulty.easy,
+    'medium': sc2_env.Difficulty.medium,
+    'medium_hard': sc2_env.Difficulty.medium_hard,
+    'hard': sc2_env.Difficulty.hard,
+    'harder': sc2_env.Difficulty.harder,
+    'very_hard': sc2_env.Difficulty.very_hard,
+    'cheat_vision': sc2_env.Difficulty.cheat_vision,
+    'cheat_money': sc2_env.Difficulty.cheat_money,
+    'cheat_insane': sc2_env.Difficulty.cheat_insane}
```
alvarofpp: What is the difference between using `sc2_difficulties` and `sc2_env.Difficulty`?

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,134 @@
...
+
+def get_sc2_difficulty(sc2_difficulty: str):
+    out = sc2_difficulties.get(sc2_difficulty)
+    if out is not None:
+        return out
+    else:
+        raise Exception("Chosen difficulty for StarCraft II doesn't match any known "
+                        "difficulties. Try: 'very_easy', 'easy', 'medium', "
+                        "'medium_hard', 'hard', 'harder', 'very_hard', "
+                        "'cheat_vision', 'cheat_money' or 'cheat_insane'")
```
alvarofpp: These functions can be avoided by allowing the developer to use `pysc2`'s own constants and functions.

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,134 @@
...
+            map_name='Simple64',
+            players=None,
+            player_race='terran',
+            enemy_race='random',
+            difficulty='very_easy',
+            render=False,
+            reset_done=True,
+            spatial_dim=16,
+            step_mul=16,
+            game_steps_per_ep=0,
```
alvarofpp: What's `ep`?

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,134 @@
...
+    'cheat_vision': sc2_env.Difficulty.cheat_vision,
+    'cheat_money': sc2_env.Difficulty.cheat_money,
+    'cheat_insane': sc2_env.Difficulty.cheat_insane}
+
+
+class SC2Env(EnvironmentBase):
+    def __init__(
+            self,
+            map_name='Simple64',
+            players=None,
```
alvarofpp: What's the point of assigning `players` here when it will be overwritten just below?

alvarofpp (changes requested):
All the code has comments of the type `TODO` or indicating some improvement. This is the time to resolve these issues and remove these comments.

RickFqt:
> All the code has comments of the type `TODO` or indicating some improvement. This is the time to resolve these issues and remove these comments.

To be honest, I put these `TODO` there to check on them later. 

For the first one on the `start()` function, I thought it would be a good idea to let the user define the parameters used in the definition of `agent_interface_format`. In that case, the user should provide the parameters from the `init` function, right?

The second `TODO` was a reafactoring issue. In the original code, those lines were like this:
```python
setattr(obs, 'step_mul', self.step_mul)
setattr(obs, 'map_size', self.env_instance._interface_formats[0]._raw_resolution)
```
But when I ran the linter code, it discouraged the use of `setattr`, and that it would be better to do a simple assignment. After that, I still wasn't able to test if these attributes of "obs" really exist, and I am still struggling on how to check that out.

urnai/sc2/environments/sc2environment.py
RickFqt: It's an episode. This parameter is used to initialize the `pysc2` environment. Should I rename it to `game_steps_per_episode`?

urnai/sc2/environments/sc2environment.py
RickFqt: I think the only difference would be allowing the user to give strings as parameters to initialize the environment. Should we use `sc2_env.Difficulty` directly?

urnai/sc2/environments/sc2environment.py
RickFqt: Actually, this parameter is not even used, as well as `obs_features`. I will remove them.

urnai/sc2/environments/sc2environment.py
alvarofpp: > Should we use sc2_env.Difficulty directly?

Yep.

urnai/sc2/environments/sc2environment.py
alvarofpp: > Should I rename it to game_steps_per_episode?

Yep.

alvarofpp:
@RickFqt 

> In that case, the user should provide the parameters from the init function, right?

It can be in `__init__` or even in `start`.

urnai/sc2/environments/sc2environment.py
```diff
@@ -4,50 +4,37 @@
...
 
 class SC2Env(EnvironmentBase):
     def __init__(
             self,
             map_name='Simple64',
-            players=None,
-            player_race='terran',
-            enemy_race='random',
-            difficulty='very_easy',
+            player_race=sc2_env.Race.terran,
```
alvarofpp: By default it could be `random` like `enemy_race`.

urnai/sc2/environments/sc2environment.py
```diff
@@ -61,26 +48,33 @@ def __init__(
...
+            raw_resolution=64,
+            use_feature_units=True,
+            screen=64,
+            minimap=64,
+    ):
         self.done = False
         if self.env_instance is None:
             self.env_instance = sc2_env.SC2Env(
                 map_name=self.id,
                 visualize=self.render,
```
alvarofpp: Why not use the same names for `self.*`?

urnai/sc2/environments/sc2environment.py
```diff
@@ -61,26 +48,33 @@ def __init__(
...
 
-    def start(self):
+    def start(
+            self,
+            action_space=actions.ActionSpace.RAW,
+            use_raw_units=True,
+            raw_resolution=64,
+            use_feature_units=True,
+            screen=64,
+            minimap=64,
```
alvarofpp: The default parameter values are redundant. I recommend either setting everything to `None` and checking during its assignment in `sc2_env.SC2Env` or change `start` to `_start`.

alvarofpp (changes requested):
It would be good to use type hint in the functions.

urnai/sc2/environments/sc2environment.py
RickFqt: That would require changing the names of attributes in [environment_base](https://github.com/UFRN-URNAI/urnai-tools/blob/issue-83/urnai/environments/environment_base.py#L15). Is that fine?

RickFqt:
> It would be good to use type hint in the functions.

Is it necessary even in functions with default parameters?

urnai/sc2/environments/sc2environment.py
RickFqt: > The default parameter values are redundant. I recommend either setting everything to `None` and checking during its assignment in `sc2_env.SC2Env` or change `start` to `_start`.

In the first suggestion, would that require an `if-else` statement in every parameter of `sc2_env.SC2Env`? 
And in the second, changing `start` to `_start` would make the method private, only being callable from the `__init__` function, right? Is it necessary for the user to be able to call `start` directly?

RickFqt:
> It would be good to use type hint in the functions.

I've been trying to find the type of the `action` parameter from the `step()` method, but I'm having some difficulty to do so. In this method, the `pysc2` own method `step()` is called with the same `action` parameter, so I started looking at its documentation [here](https://github.com/google-deepmind/pysc2/blob/master/docs/environment.md#actions). At first, I thought that an action would be an instance of [FunctionCall](https://github.com/google-deepmind/pysc2/blob/master/pysc2/lib/actions.py#L1767), since its what is returned from the [RandomAgent](https://github.com/google-deepmind/pysc2/blob/master/pysc2/agents/random_agent.py) implementation from `pysc2`. However, when trying to test this method, I wasn't able to instantiate a FunctionCall properly, and kept getting an error. This is how I tried to test the method:

```python
from pysc2.lib import actions

from urnai.sc2.environments.sc2environment import SC2Env

env = SC2Env()
obs = env.reset()

# Trying to create a function call with ID 2 (select_point)
# and ARGS = [[0], [23, 38]]
env.step(actions.FunctionCall(2, [[0], [23, 38]]))
```

That said, I'm not sure if I'm creating a `FunctionCall` object incorrectly, or if it's not the right parameter type for `step`. I also tried to look for the previous implementation of urnai, but got confused as well.

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,115 @@
...
+                    feature_dimensions=features.Dimensions(screen, minimap),
+                ),
+                step_mul=self.step_mul,
+                game_steps_per_episode=self.game_steps_per_episode,
+                realtime=self.realtime,
+            )
+
+    def step(self, action: actions.FunctionCall) -> Tuple[List[Any], int, bool]:
+        timestep = self.env_instance.step(action)
+        obs, reward, done = self.parse_timestep(timestep)
```
RickFqt: It was told in today's meeting that `done` needs to be changed to two different booleans: `terminated`, which determines if the objective was achieved, and `truncated`, which tells if the maximum number of steps was reached.

urnai/sc2/environments/sc2environment.py
alvarofpp: That's fine.

urnai/sc2/environments/sc2environment.py
alvarofpp: I think I've got it wrong. `start` will be called both in `__init__` and by the developer, correct? If so, you can leave it as it is.

alvarofpp:
> Is it necessary even in functions with default parameters?

Yes, type hint also helps IDEs when it comes to displaying function signatures.

alvarofpp:
> > It would be good to use type hint in the functions.
> 
> I've been trying to find the type of the `action` parameter from the `step()` method, but I'm having some difficulty to do so. In this method, the `pysc2` own method `step()` is called with the same `action` parameter, so I started looking at its documentation [here](https://github.com/google-deepmind/pysc2/blob/master/docs/environment.md#actions). At first, I thought that an action would be an instance of [FunctionCall](https://github.com/google-deepmind/pysc2/blob/master/pysc2/lib/actions.py#L1767), since its what is returned from the [RandomAgent](https://github.com/google-deepmind/pysc2/blob/master/pysc2/agents/random_agent.py) implementation from `pysc2`. However, when trying to test this method, I wasn't able to instantiate a FunctionCall properly, and kept getting an error. This is how I tried to test the method:
> 
> ```python
> from pysc2.lib import actions
> 
> from urnai.sc2.environments.sc2environment import SC2Env
> 
> env = SC2Env()
> obs = env.reset()
> 
> # Trying to create a function call with ID 2 (select_point)
> # and ARGS = [[0], [23, 38]]
> env.step(actions.FunctionCall(2, [[0], [23, 38]]))
> ```
> 
> That said, I'm not sure if I'm creating a `FunctionCall` object incorrectly, or if it's not the right parameter type for `step`. I also tried to look for the previous implementation of urnai, but got confused as well.

For functions, you can use the type hint `typing.Callable`.

urnai/sc2/environments/sc2environment.py
alvarofpp: I don't think these are booleans, but statuses. In this case it would be better to use `Enum`, `string` or `constants`.

RickFqt:
> For functions, you can use the type hint `typing.Callable`.

I've searched a little bit and it seems like importing `typing.Callable` is deprecated, as well as some other modules from `typing` (see [here](https://peps.python.org/pep-0585/#implementation)). Is that right? If so, should I change the other things imported from typing like `typing.List` or `typing.Tuple`?

alvarofpp:
> > For functions, you can use the type hint `typing.Callable`.
> 
> I've searched a little bit and it seems like importing `typing.Callable` is deprecated, as well as some other modules from `typing` (see [here](https://peps.python.org/pep-0585/#implementation)). Is that right? If so, should I change the other things imported from typing like `typing.List` or `typing.Tuple`?

That's right, I didn't know that. We will adopt the imports suggested by PEP from now on.

urnai/environments/environment_base.py
```diff
@@ -11,9 +10,9 @@ class EnvironmentBase(ABC):
     should still be used as a wrapper (e.g. implementing an environment for OpenAI gym).
     """
 
-    def __init__(self, id: str, render=False, reset_done=True):
-        self.id = id
-        self.render = render
+    def __init__(self, map_name: str, visualize=False, reset_done=True):
```
alvarofpp:
```suggestion
    def __init__(self, map_name: str, visualize: bool = False, reset_done: bool = True):
```

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.self_play:
+            self.players = [
+                sc2_env.Agent(self.player_race),
+                sc2_env.Agent(self.enemy_race),
+            ]
+        else:
+            self.players = [
+                sc2_env.Agent(self.player_race),
+                sc2_env.Bot(self.enemy_race, self.difficulty),
+            ]
```
alvarofpp: What if I want to use more than two races?

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,123 @@
...
+            use_raw_units=True,
+            raw_resolution=64,
+            use_feature_units=True,
+            screen=64,
+            minimap=64,
+    ):
+        super().__init__(map_name, visualize, reset_done)
+
+        FLAGS = flags.FLAGS
+        FLAGS(sys.argv)
```
alvarofpp:
```suggestion
       flags.FLAGS(sys.argv)
```

alvarofpp (changes requested):
All that remains is to remove the comments and insert the unit tests.

urnai/sc2/environments/sc2environment.py
RickFqt: Do you mean more than two races in a single game? Like Zergs vs Terran vs Protoss (3 players or more against each other)?

urnai/sc2/environments/sc2environment.py
RickFqt: I forgot to ask when I put this part of the code, but do you know what is the purpose of it? I tried to run the program without it and kept getting errors, but when I put these flags it worked normally. It was on the previous version of urnai, so I just copied it from there.

urnai/environments/environment_base.py
RickFqt: I didn't know this could be done with default parameters :0
Should I do it with the other ones from other methods too?

RickFqt:
> All that remains is to remove the comments and insert the unit tests.

About the unit tests, I have some questions:

- Do we have to test the functionality of other classes that are used inside the class? For example, here in our `SC2Env` class, we have an attribute called `env_instance`, which is initialized with an instance of another class (`sc2_env.SC2Env`). When a method of `SC2Env` calls methods from `env_instance`, do we have to check the behavior of attributes that are inside `env_instance`?
- When testing the method `step`, I was using the previous implementation of urnai's [action wrapper](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2_wrapper.py) to create instances of actions, and then passed them as a parameter to `step` to test the functionality of some actions (like build refinery, for example). Is there a way of not using this and still simulate the functionality of the actions?

urnai/environments/environment_base.py
alvarofpp: Yes!

urnai/sc2/environments/sc2environment.py
alvarofpp: Both.

urnai/sc2/environments/sc2environment.py
alvarofpp: I don't know what that part of the code is for. I recommend removing it and fixing any errors that appear.

alvarofpp:
> - Do we have to test the functionality of other classes that are used inside the class? For example, here in our `SC2Env` class, we have an attribute called `env_instance`, which is initialized with an instance of another class (`sc2_env.SC2Env`). When a method of `SC2Env` calls methods from `env_instance`, do we have to check the behavior of attributes that are inside `env_instance`?

In the case of unit tests, we don't need to test the classes of third-party packages or the classes used by the class we're testing. The recommended thing is to create mocks for these other classes and simulate the expected behavior.

> - When testing the method `step`, I was using the previous implementation of urnai's [action wrapper](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2_wrapper.py) to create instances of actions, and then passed them as a parameter to `step` to test the functionality of some actions (like build refinery, for example). Is there a way of not using this and still simulate the functionality of the actions?

I recommend that you only use the old code as a reference and not as a test. For the tests, try to stick to the code we currently have in `v2`.

You can and should use mocks to simulate the expected behavior (as you did with the fake classes in another issue).

urnai/sc2/environments/sc2environment.py
RickFqt: I searched a bit into the [pysc2 code implementation](https://github.com/google-deepmind/pysc2/blob/master/pysc2/env/sc2_env.py#L184) of their environment, and apparently they only have support to at most 2 players on a single map. Here, `players` is the list of Agents or Bots passed as a parameter to their initialization method:

```python
for p in players:
  if not isinstance(p, (Agent, Bot)):
    raise ValueError(
        "Expected players to be of type Agent or Bot. Got: %s." % p)

num_players = len(players)
self._num_agents = sum(1 for p in players if isinstance(p, Agent))
self._players = players

if not 1 <= num_players <= 2 or not self._num_agents:
  raise ValueError("Only 1 or 2 players with at least one agent is "
                    "supported at the moment.")
```

RickFqt: But even with at most 2 players at a time, our current code supports the following:
- One Agent against one Bot
- One Agent against another Agent

For the second case, should our environment return two instances of observations and rewards on the method `step()`, for example? Because I believe our current implementation only returns the observation and reward of the first Agent.

tests/units/sc2/environments/test_sc2_environment.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_reset_method(self):
+        # TODO
+        pass
+
+    @patch('pysc2.env.sc2_env.SC2Env')
+    def test_close_method(self, MockSC2Env):
+        # TODO: Check why this does not work
+        # env = SC2Env()
+        # env.close()
+        # MockSC2Env.close.assert_called_once()
```
RickFqt: I tried to test the `close()` method with this commented code, but it didn't work well. My idea was:
- In `env = SC2Env()`, when inside the method `start()`, instanciate the local variable `env_instance` with the mocked class.
- In `env.close()`, call the `close()` method from `env_instance` (which was instanciated with the mocked class).
- Check if the method `close()` from the mocked class was called only once.

But when testing it out, the test fails, asserting that the method `close()` from the mocked class was never called.

RickFqt: I think it worked out. Now the code is like this, and the tests run normally:

```python
@patch('pysc2.env.sc2_env.SC2Env')
    def test_close_method(self, MockSC2Env):
        env = SC2Env()
        env.close()
        env.env_instance.close.assert_called_once()
```

Is that correct?

urnai/sc2/environments/sc2environment.py
alvarofpp: What a sad thing to allow only 2 agents. I don't know how pysc2 will return the observations in this case, I believe the agents should have an identifier to differentiate them. For now, let's just focus on one Agent and later we can implement two Agents. @RickFqt create a ticket about this, please.

tests/units/sc2/environments/test_sc2_environment.py
alvarofpp: I don't think that's right. I think the correct thing would be something like:

```suggestion
    @patch('pysc2.env.sc2_env.SC2Env')
    def test_close_method(self, scenv_mock):
        # GIVEN
        env = SC2Env()
        # WHEN
        env.close()
        # THEN
        scenv_mock.env_instance.close.assert_called_once()
```

alvarofpp: I recommend using `given-when-then` when writing your tests. All you have to do is separate your test with the 3 comments I've shown above. Here's a link if you want to dig deeper: [Given-When-Then](https://pythontest.com/strategy/given-when-then-2/)

urnai/sc2/environments/sc2environment.py
RickFqt: > @RickFqt create a ticket about this, please.

Do you mean creating a new issue?

tests/units/sc2/environments/test_sc2_environment.py
RickFqt: > I don't think that's right. I think the correct thing would be something like:

I tried this code and got the same error as before, `close` is never called.

urnai/sc2/environments/sc2environment.py
alvarofpp: Yes.

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,112 @@
...
+        super().__init__(map_name, visualize, reset_done)
+
+        # TODO: Investigar como remover essa dependência
+        flags.FLAGS(sys.argv)
+
+        self.step_mul = step_mul
+        self.game_steps_per_episode = game_steps_per_episode
+        self.spatial_dim = spatial_dim
+        if(players is None):
+            # Default value for players
```
alvarofpp, line 38: Remove comment.

urnai/sc2/environments/sc2environment.py
```diff
@@ -0,0 +1,112 @@
...
+    def step(self, action: list) -> tuple[list, int, bool, bool]:
+        timestep = self.env_instance.step(action)
+        obs, reward, terminated, truncated = self._parse_timestep(timestep)
+        self.terminated = terminated
+        self.truncated = truncated
+        return obs, reward, terminated, truncated
+
+    def reset(self) -> list:
+        timestep = self.env_instance.reset()
+        obs, _, _, _ = self._parse_timestep(timestep)
```
alvarofpp, line 86:
```suggestion
        obs, *_ = self._parse_timestep(timestep)
```

alvarofpp (approved):

PR #88, opened 2024-06-07
Resolves #68

### LINKED ISSUE #68: Create base class for action space
We need a class that represents an agent's action space. This class must have a set of actions and allow the agent to know which actions are possible or not.

This issue depends on issue #67 being completed first.

Suggestions:

- Class name: `ActionSpaceBase`.
- Class path: `urnai.actions`.
- Source: https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/base/abwrapper.py .

Notes:

CinquilCinquil:


alvarofpp:
Please rebase your branch.

CinquilCinquil:
I tried, but it apparently isn't recognizing that the 15 commits i did in issue-67 are already in the v2. I think its because those commits are all squashed inside a commit called ["Resolves #67 (#79)"](https://github.com/UFRN-URNAI/urnai-tools/commit/e461f3ebeb3daf7de03ebbaf8c5b8ab1fb83a275). So when i try to rebase, it asks me to solve all of them manually (wich i don't think is the right way), like its treating them as totally new commits, despite the fact that they arrive at the same end result.

What i thought about doing was saving my changes locally, then deleting this branch and creating it again with the right base, after that commit my changes again, since they weren't that big. What do you think @alvarofpp ?

alvarofpp:
@CinquilCinquil 

1. Close this PR (#88);
2. Create a new branch from the `v2` branch;
3. Run `cherry-pick` for your commits in the `issue-68` branch;
4. Open a new PR with the changes from the branch created in step 2.

PR #89, opened 2024-06-14
Resolves #68

### LINKED ISSUE #68: Create base class for action space

CinquilCinquil:
#68

### DIFF
--- a/tests/units/actions/test_action_base.py
+++ b/tests/units/actions/test_action_base.py
@@ -7,19 +7,23 @@ from urnai.actions.action_base import ActionBase
 class FakeAction(ActionBase):
     ActionBase.__abstractmethods__ = set()
     __id__ = None
+
+    def __init__(self, name):
+        self.name = name
     ...
 
 class TestActionBase(unittest.TestCase):
 
     def test_abstract_methods(self):
 
-        fake_action = FakeAction()
+        fake_action = FakeAction("test_name")
         run_return = fake_action.run()
         check_return = fake_action.check("observation")
-        is_complete_return = fake_action.is_complete()
+        is_complete_return = fake_action.is_complete
+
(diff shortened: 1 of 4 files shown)


tests/units/actions/test_action_base.py
```diff
@@ -7,19 +7,23 @@
 class FakeAction(ActionBase):
     ActionBase.__abstractmethods__ = set()
     __id__ = None
+
+    def __init__(self, name):
+        self.name = name
```
alvarofpp: Why is that if you already have the id? Besides, `ActionBase` doesn't have `name` in `__init__`.

alvarofpp (changes requested):

tests/units/actions/test_action_base.py
CinquilCinquil: There was a `get_named_actions` method in the reference code for this class and i found it interesting for debugging pourposes. Then, i thought the name and the id were separete things, like one for the user and the other for the system. 

I had put the name attribute in `ActionBase` but i think it was lost when i changed to this branch.

alvarofpp: If I remember correctly, the `id` would also be defined by the user. So it ends up that `id` and `name` have the same purpose, making it redundant to have them. I think we can remove `name`.

alvarofpp (approved):

PR #90, opened 2024-06-14
Resolves #87

### LINKED ISSUE #87: Add support for two Agents on a single sc2 environment
Our current environment implementation only gives support for one Agent per environment. We need to add support for two Agents on a single environment.

- Note: Currently, pysc2 only gives support to [at most 2 players on a single map](https://github.com/google-deepmind/pysc2/blob/master/pysc2/env/sc2_env.py#L184). When this limit is expanded, it would be a good thing to support multiple agents on a single map.

RickFqt:
Resolves #87.

- [X] Updates `sc2environment` class, so that it supports more than 1 agent per environment.
- [X] Refactors the unit tests accordingly to the changes made.

### DIFF
--- a/urnai/sc2/environments/sc2environment.py
+++ b/urnai/sc2/environments/sc2environment.py
@@ -35,12 +35,12 @@ class SC2Env(EnvironmentBase):
         self.game_steps_per_episode = game_steps_per_episode
         self.spatial_dim = spatial_dim
         if(players is None):
-            # Default value for players
             players = [
                 sc2_env.Agent(sc2_env.Race.random),
                 sc2_env.Bot(sc2_env.Race.random, sc2_env.Difficulty.very_easy),
             ]
         self.players = players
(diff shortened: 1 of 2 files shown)


urnai/sc2/environments/sc2environment.py
```diff
@@ -93,20 +93,27 @@ def restart(self) -> None:
...
-        first_timestep = timestep[0]
-        obs, reward =  first_timestep.observation,  first_timestep.reward
+        obs = []
+        reward = []
+        map_size = self.env_instance._interface_formats[0]._raw_resolution
+        
+        for agent_index in range(self.num_agents):
+            obs.append(timestep[agent_index].observation)
+            reward.append(timestep[agent_index].reward)
+            # add step_mul and map_size to obs
```
alvarofpp: Remove comment.

alvarofpp (changes requested):

alvarofpp (approved):

PR #91, opened 2024-06-14
Resolves #75

### LINKED ISSUE #75: Create base class for models

CinquilCinquil:
The v2 branch doesn't seem to be using 'utils' so the `urnai.utils` reference in the code was causing an error. #75

### DIFF
--- a/urnai/models/model_base.py
+++ b/urnai/models/model_base.py
@@ -1,8 +1,6 @@
 from abc import ABC, abstractmethod
 from typing import Dict
 
-from urnai.utils.returns import ActionIndex
-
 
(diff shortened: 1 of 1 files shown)


alvarofpp (approved):

alvarofpp:
@CinquilCinquil A tip: next time don't reopen an issue that has already been closed. Instead, create a new issue and reference the old issue if necessary, most of the time you won't need to reference the old issue.

PR #92, opened 2024-06-21
Resolves #76

CinquilCinquil:
#76

Issue #74, opened 2023-11-23
Create action choice strategy class


alvarofpp:
After issues #67 and #68, we will have a set of classes that will allow us to work with agent actions. This class should have a `choose` method that receives a set of actions and chooses one.

Suggestions:

- Class name: `StrategyBase`.
- Class path: `urnai.strategies`.

alvarofpp:
Importante: https://github.com/UFRN-URNAI/urnai-tools/pull/93#issuecomment-2211404541

alvarofpp:
Only close this issue after #94 is merged.

PR #94, opened 2024-07-12
Resolves #76 (2)

### LINKED ISSUE #76: Create a base class for agents
Our agents need to be given their action space, state space, and DRL model. It needs to have a way of interacting with the Environment and storing the rewards it receives for each interaction.

Suggestions:

- Class name: `AgentBase`.
- Class path: `urnai.agents`.
- Source: https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/sc2_agent.py .

CinquilCinquil:
#76 

Continuation of the pull request #92 .

### DIFF
--- a/urnai/agents/agent_base.py
+++ b/urnai/agents/agent_base.py
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 
+from urnai.actions.action_base import ActionBase
 from urnai.actions.action_space_base import ActionSpaceBase
 from urnai.models.model_base import ModelBase
 from urnai.states.state_base import StateBase
@@ -25,6 +26,13 @@ class AgentBase(ABC):
 
     @abstractmethod
     def step(self) -> None:
+        ...
+
+    @abstractmethod
+    def choose_action(self, action_space : ActionSpaceBase) -> ActionBase:
+        """
+        Method that contains the agent's strategy for choosing actions
+        """
         ...
 
     def reset(self, episode=0) -> None:
--- a/urnai/agents/agent_base.py
+++ b/urnai/agents/agent_base.py
@@ -11,18 +11,18 @@ class AgentBase(ABC):
     def __init__(self, action_space : ActionSpaceBase,
                  state_space : StateBase,
                  model : ModelBase,
-                 reward_builder):
+                 reward):
 
         self.action_space = action_space
         self.state_space = state_space
         self.model = model
-        self.reward_builder = reward_builder
+        self.reward = reward
 
         self.previous_action = None
         self.previous_state = None
 
         self.attr_block_list = ['model', 'action_space',
-                                'state_space', 'reward_builder']
+                                'state_space', 'reward']
 
     @abstractmethod
     def step(self) -> None:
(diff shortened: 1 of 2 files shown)


urnai/agents/agent_base.py
```diff
@@ -0,0 +1,70 @@
...
+from urnai.models.model_base import ModelBase
+from urnai.states.state_base import StateBase
+
+
+class AgentBase(ABC):
+
+    def __init__(self, action_space : ActionSpaceBase,
+                 state_space : StateBase,
+                 model : ModelBase,
+                 reward_builder):
```
alvarofpp:
```suggestion
                 reward):
```

urnai/agents/agent_base.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.action_space = action_space
+        self.state_space = state_space
+        self.model = model
+        self.reward_builder = reward_builder
```
alvarofpp:
```suggestion
        self.reward = reward
```

urnai/agents/agent_base.py
```diff
@@ -0,0 +1,70 @@
...
+            next_state = self.update_state(obs)
+            self.model.learn(self.previous_state, self.previous_action,
+                              reward, next_state, done)
+
+
+    def update_state(self, obs) -> list:
+        """
+        Returns the state of the game environment
+        """
+        return self.state_space.update(obs)
```
alvarofpp: I don't see much point in having a method like this, since the attributes are public and not private.

urnai/agents/agent_base.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def get_reward(self, obs, reward, done) -> None:
+        """
+        Calls the get_reward method from the reward_builder, effectivelly
+        returning the reward value.
+        """
+        return self.reward_builder.get_reward(obs, reward, done)
```
alvarofpp: I don't see much point in having a method like this, since the attributes are public and not private.

urnai/agents/agent_base.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def state_dim(self) -> int:
+        """Returns the dimensions of the state builder"""
+        return self.state_space.dimension
```
alvarofpp: I don't see much point in having a method like this, since the attributes are public and not private.

alvarofpp (changes requested):

alvarofpp:
@CinquilCinquil Since the lint has broken, my recommendation is:

- Remove the last commit.
- As there were only a few changes, make them again in your current branch and submit a commit setting the context to #74.

You will have to forcefully update your branch (`-f`).

urnai/agents/agent_base.py
```diff
@@ -0,0 +1,59 @@
...
+        self.model.ep_reset(episode)
+        self.reward.reset()
+        self.state_space.reset()
+
+    def learn(self, obs, reward, done) -> None:
+        """
+        If it is not the very first step in an episode, this method will
+        call the model's learn method.
+        """
+        if self.previous_state is not None:
```
alvarofpp, line 56: At what point is `self.previous_state` not set to `None`? From what I've seen, this is its initial “value” and it is not changed anywhere, so it will always have this value and will never enter this `if`.

CinquilCinquil: This method is ment to be used in a child class where depending on the situation `self.previous_state` might not be `None`.

alvarofpp (approved):

PR #99, opened 2024-07-30
Resolves #98

### LINKED ISSUE #98: Action class for pysc2
It was discussed in a meeting that the `Action` class should encapsulate the action behaviour from pysc2.

CinquilCinquil:
#98

### DIFF
--- a/tests/units/sc2/actions/test_sc2_action.py
+++ b/tests/units/sc2/actions/test_sc2_action.py
@@ -7,7 +7,7 @@ from urnai.actions.sc2_action import SC2Action
 _BUILD_REFINERY = actions.RAW_FUNCTIONS.Build_Refinery_pt
 _NO_OP = actions.FUNCTIONS.no_op
 
-class TestActionBase(unittest.TestCase):
+class TestSC2Action(unittest.TestCase):
 
     def test_run(self):
 
--- a/tests/units/sc2/actions/test_sc2_action.py
+++ b/tests/units/sc2/actions/test_sc2_action.py
@@ -1,0 +1,21 @@
+import unittest
+
+from pysc2.lib import actions
+
+from urnai.actions.sc2_action import SC2Action
+
+_BUILD_REFINERY = actions.RAW_FUNCTIONS.Build_Refinery_pt
+_NO_OP = actions.FUNCTIONS.no_op
+
+class TestActionBase(unittest.TestCase):
+
+    def test_run(self):
+
+        run_no_op = SC2Action.run(_NO_OP)
+        run_build_refinery = SC2Action.run(_BUILD_REFINERY, 'now', 0)
+        
+        self.assertEqual(run_no_op.function, _NO_OP.id)
+        self.assertEqual(run_no_op.arguments, [])
+
+        self.assertEqual(run_build_refinery.function, _BUILD_REFINERY.id)
+        self.assertEqual(run_build_refinery.arguments, [[actions.Queued['now']], [0]])
--- a/urnai/sc2/actions/sc2_action.py
+++ b/urnai/sc2/actions/sc2_action.py
@@ -5,5 +5,5 @@ class SC2Action:
 
     """This class encapsulates the usage of actions from pysc2"""
 
-    def run(action, *args) -> FunctionCall:
-        return action(*args)
+    def run(action_function, *args) -> FunctionCall:
+        return action_function(*args)
--- a/urnai/sc2/actions/sc2_action.py
+++ b/urnai/sc2/actions/sc2_action.py
@@ -1,0 +1,9 @@
+from pysc2.lib.actions import FunctionCall
+
+
+class SC2Action:
+
+    """This class encapsulates the usage of actions from pysc2"""
+
+    def run(action_id, *args) -> FunctionCall:
+        return action_id(*args)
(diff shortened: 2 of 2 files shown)


tests/units/sc2/actions/test_sc2_action.py
```diff
@@ -0,0 +1,21 @@
+import unittest
+
+from pysc2.lib import actions
+
+from urnai.actions.sc2_action import SC2Action
+
+_BUILD_REFINERY = actions.RAW_FUNCTIONS.Build_Refinery_pt
+_NO_OP = actions.FUNCTIONS.no_op
+
+class TestActionBase(unittest.TestCase):
```
alvarofpp:
```suggestion
class TestSC2Action(unittest.TestCase):
```

urnai/sc2/actions/sc2_action.py
```diff
@@ -0,0 +1,9 @@
+from pysc2.lib.actions import FunctionCall
+
+
+class SC2Action:
+
+    """This class encapsulates the usage of actions from pysc2"""
+
+    def run(action_id, *args) -> FunctionCall:
```
alvarofpp: It seems to me that `action_id` is a function and not an identifier. Is that correct?

alvarofpp (changes requested):

urnai/sc2/actions/sc2_action.py
CinquilCinquil: from what i understand its kind of both at the same time.

I took this name from a code from the previous version of urnai i was using to make the sc2 bot. In there, a function receives an `action_id` and then uses it as a function to generate a `FunctionCall`, wich then is passed to the environment.

I think it makes sense to leave it as `action_id` because thats what the user will be passing into the function.

CinquilCinquil: I think it makes sense to leave it as `action_id` because thats what the user will be passing into the function.

alvarofpp: Can you give an example of this? I can understand the logic of saying that a function is an identifier, because we won't have the same functions in the same scope, but following the same logic classes are also identifiers. Furthermore, this logic ignores what we have as a definition of what an identifier is. The callback itself is `FunctionCall`, which implies that the `action_id` is a function and not an identifier.

CinquilCinquil: An example is present in the function `build_structure_by_type` in [sc2.py](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2.py) (more precisely in line 143)

alvarofpp: Okay, if you follow the invocations up to line 143, you'll see that `action_id` is a function. This is in addition to verifying that `action_id` is in fact a function just because it has an invocation (`action(...)`) and a return is expected.

Let's do the trace:
- On line [143 of the sc2.py](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2.py#L143) file we have an invocation of `action_id`. This code is part of the `build_structure_by_type` function defined on line [132](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2.py#L132C5-L132C28) of the same file.
- One of the invocations of `build_structure_by_type` occurs on line [427](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2.py#L427) of the same file (`sc2.py`). This in turn is part of the `build_structure_raw` function defined on line [420](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2.py#L420) in the same file.
- One of the `build_structure_raw` invocations occurs on line [949 of the sc2_wrapper.py](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2_wrapper.py#L949) file. Here we can see that the third argument passed is `sc2._BUILD_TECHLAB_BARRACKS`.
- If we go back to the [`sc2.py` file, on line 42](https://github.com/UFRN-URNAI/urnai-tools/blob/main/urnai/agents/actions/sc2.py#L42) we have the definition of `_BUILD_TECHLAB_BARRACKS` which consists of `actions.RAW_FUNCTIONS.Build_TechLab_Barracks_quick`.
- If we go into [pysc2](https://github.com/google-deepmind/pysc2) and check where `actions.RAW_FUNCTIONS.Build_TechLab_Barracks_quick` is defined, we'll see that it's on [line 1270 of the `actions.py` file](https://github.com/google-deepmind/pysc2/blob/master/pysc2/lib/actions.py#L1270).
- Checking the [signature of the `raw_ability` function](https://github.com/google-deepmind/pysc2/blob/master/pysc2/lib/actions.py#L497C7-L497C18), we can see that this is the creation of a function that represents an action in the game.

Having said all that, I disagree that this is an ID, but rather a function. The old code was wrong to use this nomenclature and I don't think we should repeat this mistake.

urnai/sc2/actions/sc2_action.py
```diff
@@ -0,0 +1,9 @@
+from pysc2.lib.actions import FunctionCall
+
+
+class SC2Action:
+
+    """This class encapsulates the usage of actions from pysc2"""
+
+    def run(action, *args) -> FunctionCall:
+        return action(*args)
```
alvarofpp:
```suggestion
    def run(action_function, *args) -> FunctionCall:
        return action_function(*args)
```

alvarofpp (changes requested):

alvarofpp (approved):

PR #102, opened 2024-08-12
Implementing a bot for sc2 collectables minigame

RickFqt:
When this PR is complete, it should implement a bot that can be trained and used to solve the starcraft2 collectables minigame.

TODO list:
- [x] Fix linter checks
- [x] Implement the persistance class
- [x] Fix the method `choose_action` from `agent_base` and `model_base`
- [ ] Fix existing tests

### DIFF
--- a/sc2_collectables_trainer.py
+++ b/sc2_collectables_trainer.py
@@ -1,0 +1,48 @@
+import pathlib
+import sys
+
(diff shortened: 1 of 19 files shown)


CinquilCinquil:
I had to adapt the Persistence class a little bit for this issue, and i think the Persistence class in v2 should also suffer these changes. Basically, it now receives a `object_to_save` and stores/loads the data onto it instead of using the class itself like before.

PR #106, opened 2024-10-08
Achieve 100% test coverage for agent_base class

RickFqt:
This PR should:

- [X] Finish the unit tests for `agent_base` class to achieve 100% test coverage.

### DIFF
--- a/urnai/models/model_base.py
+++ b/urnai/models/model_base.py
@@ -10,7 +10,7 @@ class ModelBase(ABC):
 		self.learning_data = {}
 	
 	@abstractmethod
-	def learn(self, current_state, action, reward, next_state):
+	def learn(self, current_state, action, reward, next_state, done):
 		"""Learning strategy"""
 		...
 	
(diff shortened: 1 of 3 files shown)


urnai/models/model_base.py
```diff
@@ -10,7 +10,7 @@ def __init__(self):
 		self.learning_data = {}
 	
 	@abstractmethod
-	def learn(self, current_state, action, reward, next_state):
+	def learn(self, current_state, action, reward, next_state, done):
```
RickFqt, line 13: When testing the `learn` method from `agent_base` class with a `previous_state` set, the method calls the `learn` method from `model_base`, and passed an extra parameter `done` to it. So, I had to add this parameter to the original abstract method from `model_base` class, and had to add it to the method that tests `model_base` too. (`test_model_base`)

Another way to solve this would be removing `done` from the call made inside agent_base's `learn` method, but I think that this paramater should be important in the future when implementing the subclasses for it.

alvarofpp (approved):

PR #114, opened 2024-10-27
Resolve #111 Add test for PersistencePickle._simple_save

### LINKED ISSUE #111: Find out why a Mock doesn't cover the unit tests in PersistencePickle
The command `coverage report -m` shows that the lines 33-38 in the [PersistencePickle ](https://github.com/UFRN-URNAI/urnai-tools/blob/v2/urnai/base/persistence_pickle.py)class are not covered by tests, even though tests for that part [exist](https://github.com/UFRN-URNAI/urnai-tools/blob/v2/tests/units/base/test_persistence_pickle.py). The purpouse of this issue is to find out why this happens and implement a solution.

image

alvarofpp:
- Upgrade `mambaorg/micromamba` Docker image base from `1.4.9` to `2.0.8`.
- Downgrade Python from `3.11` to `3.10` to fix pysc2 incompatibility (`TypeError: Random.shuffle() takes 2 positional arguments but 3 were given` on `pysc2/lib/colors.py:121`).
- Add test for `PersistencePickle._simple_save`.
- Decrease `coverage.fail_under` from `100.0` to `70.0`.

Related:
- #111
- #122 (maybe?)

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1,41 +1,74 @@
 ---
 version: '3'
 
-env:
+vars:
+  LINTER_COMMIT_TARGET_BRANCH: origin/v2
+  LINTER_DOCKER_IMAGE: urnai/linter:latest
   PACKAGE_NAME: urnai
-  DOCKER_LINTER_IMAGE: urnai/linter:latest
-  LINT_COMMIT_TARGET_BRANCH: origin/v2
   TEST_CONTAINER_NAME: urnai_test
-  ROOT:
-    sh: pwd
-  CMD_DOCKER_RUN: docker run --rm
   VIRTUAL_ENVIRONMENT: urnai_env
 
+env:
+  CMD_DOCKER_RUN: docker run --rm
+  ROOT:
+    sh: pwd
+  UID:
+    sh: id -u
+  GID:
+    sh: id -g
+
 tasks:
+  default:
+    desc: "List all tasks."
+    cmds:
+      - task --list-all
+
   build:
     desc: Builds the Docker image.
-    deps: [ config-hooks ]
+    summary: |
+      Builds the Docker image.
+
+      Examples:
+        - task build
+    deps:
+      - config-hooks
     cmds:
-      - docker build -t ${PACKAGE_NAME} .
+      - docker build -t {{ .PACKAGE_NAME }} .
 
   build-no-cache:
     desc: Builds the Docker image without using cache.
-    deps: [ config-hooks ]
+    summary: |
+      Builds the Docker image without using cache.
+
+      Examples:
+        - task build-no-cache
+    deps:
+      - config-hooks
     cmds:
-      - docker build --no-cache -t ${PACKAGE_NAME} .
+      - docker build --no-cache -t {{ .PACKAGE_NAME }} .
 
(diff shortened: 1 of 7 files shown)


Taskfile.yml
RickFqt, line 1: I tried to run `task shell` and the `task test` related commands and got the following errors:

image

Taskfile.yml
```diff
@@ -50,17 +50,23 @@ tasks:
   shell:
     desc: Runs the shell of a container.
     cmds:
-      - ${CMD_DOCKER_RUN} -v ${ROOT}:/tmp -it ${PACKAGE_NAME} /bin/bash
+      - ${CMD_DOCKER_RUN} -u ${UID}:${GID -v ${ROOT}:/tmp -it ${PACKAGE_NAME} /bin/bash
```
RickFqt:
```suggestion
      - ${CMD_DOCKER_RUN} -u ${UID}:${GID} -v ${ROOT}:/tmp -it ${PACKAGE_NAME} /bin/bash
```

Taskfile.yml
CinquilCinquil: I'm getting the same errors and after fixing the `task shell` one it also outputs "uids and gids must be in range 0-2147483647". (i'm on windows btw)

alvarofpp: @RickFqt @CinquilCinquil

This error occurs precisely because you use Windows. Docker images are usually based on Linux and usually have Mac compatibility (because of the relationship between Linux, Unix, FreeBSD, but that's another story), but for some things Windows doesn't work. In fact, there are many reasons why Windows is not a good OS for working with software development. The vast majority of people who answer that they use Windows for software development are actually referring to WSL. While for Linux/Mac the user and group identifiers are integers (like `1000`), on Windows it's for example “S-1-5-21-3555162687-1187815339-3413526530-1002”.

I've removed the dependency on specifying user for most commands, but I can't do it for some (like `test-coverage` because it updates `.coverage`).

Perhaps it would be worthwhile for one of you two (who use Windows) to investigate this a little more and find an alternative, as you already have the project environment set up on your machines.

CinquilCinquil:
@alvarofpp All changes unrelated to lint are ok! But, my local lint still complains about 2 errors (`experiments/solves/solve_collectables_sb3.py:8:1: I001 [*] Import block is un-sorted or un-formatted` and `urnai/trainers/stablebaselines3_trainer.py:1:1: I001 [*] Import block is un-sorted or un-formatted`) which the github lint does not mention

alvarofpp:
@CinquilCinquil 

> @alvarofpp All changes unrelated to lint are ok! But, my local lint still complains about 2 errors (`experiments/solves/solve_collectables_sb3.py:8:1: I001 [*] Import block is un-sorted or un-formatted` and `urnai/trainers/stablebaselines3_trainer.py:1:1: I001 [*] Import block is un-sorted or un-formatted`) which the github lint does not mention

I'll sort this out in #122. Can you put the following information in the issue please?

- Branch of your code;
- Command you are using;
- These two errors that are still showing.

CinquilCinquil (approved):

RickFqt (approved):

PR #119, opened 2025-03-25
Remove outdated SC2Action

CinquilCinquil:
After a recent merge, there were still parts of the code that utilized the old class `SC2Action` for actions.

### DIFF
--- a/experiments/solves/solve_collectables_sb3.py
+++ b/experiments/solves/solve_collectables_sb3.py
@@ -5,20 +5,20 @@ import numpy as np
 
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..')))
 
+import wandb
 from absl import app
 from gymnasium import spaces
 from pysc2.env import sc2_env
 from stable_baselines3 import PPO
 from stable_baselines3.common.monitor import Monitor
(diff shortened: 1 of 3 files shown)


alvarofpp:
@CinquilCinquil you need to apply the lint

CinquilCinquil:
yes, the only failed check was `urnai/sc2/actions/collectables.py:1:1: I001 [*] Import block is un-sorted or un-formatted`. I didn't manage to fix it manually and the `--fix` flag doesn't work. We had already discussed this and another solution was proposed a while back, but i tried that and it didn't work either

In summary, i can't solve these import issues.

CinquilCinquil:
Also, there is a discrepancy between what my local lint warns me and what the GitHub one does. On my local one, both this check and another import related one in `urnai/trainers/stablebaselines3_trainer.py` are failing, but here only the first one is.

CinquilCinquil:
@alvarofpp can you help me solve these problems or at least, for now, use lint's `--fix` flag on your computer?

alvarofpp:
@CinquilCinquil I've fixed it to pass lint. But fix this problem on your machine so that it doesn't block you in the future. How are you running lint on your machine?

alvarofpp (approved):

PR #128, opened 2025-08-14
Encountered bug when loading saved model schedule with singleton

CinquilCinquil:
## Context

My task was to create a schedule for the `clip_range` parameter of PPO (or MaskablePPO in this branch) that varied with the number of **iterations** of training **sessions**. However, Stable baselines only allowed me to create a schedule that had access to the progress of a single training **session**, which would always reset in a new **iteration**.

Therefore, i had to work with some sort of global variable to make it work. In this pull request, i tried the Singleton pattern (as described [here](https://refactoring.guru/design-patterns/singleton/python/example#example-0)), and it worked at first! But i encountered a problem when loading a saved model that already had undergone at least one iteration.

## The bug

The `id` of the instance used inside the schedule is different from the one created when the program was executed after i load, consequently their variables differ, which unables me to update the schedule correctly. This happens even though they are both instances of the same singleton. So, i'm looking for suggestions on how to tackle this problem.

## Execution

The execution can be done with `task run-ui -- experiments/solves/solve_buildmarines_sb3.py` . And i've altered the code to run the experiments faster (it will run for 2 iterations) and included prints needed for debugging.

(Disclaimer: the intent of this pull request is only to solve this bug, the code presented is not polished or finished)


CinquilCinquil:
The bug was fixed by relying on a schedule with an internal memory. This solution was chosen due to both global variables and singletons failing to update the values utilized by the schedule, after loading it from a saved model, with the ones of the program execution.

Warning: This method does not guarantee that the internal iteration value of the schedule matches the external one.

PR #129, opened 2025-09-08
Deep Mind State class

RickFqt:
When complete, this PR should try to implement the same State representation used by DeepMind on the _Atari-net Agent_, proposed on their [article](https://arxiv.org/pdf/1708.04782).

To improve visualization, these pictures give an example of the transformation of the observation into the desired state:

- Non-spatial features: apply a numerical re-escale into every element;
image

- Minimap feature layers: separate categorical layers from scalar layers. On the categorical, apply one-hot encoding, followed by 1x1 convolution. On the scalar, apply the numerical re-escale;
image

- Screen feature layers: same processing as minimap feature layers.
image

### DIFF
--- a/urnai/sc2/states/deepmind_state.py
+++ b/urnai/sc2/states/deepmind_state.py
@@ -11,7 +11,7 @@ from urnai.states.state_base import StateBase
 class DeepmindState(StateBase):
     """State processor used by DeepMind for StarCraft II observations."""
 
-    def __init__(self, categorical_conv_channels: int = 8, 
+    def __init__(self, categorical_conv_channels: int = 1, 
                  normalize_data: bool = False):
         self.categorical_out_conv_channels = categorical_conv_channels
         self.normalize_data = normalize_data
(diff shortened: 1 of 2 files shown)


urnai/sc2/states/deepmind_state.py
```diff
@@ -0,0 +1,217 @@
...
+        self.reset()
+
+    def reset(self):
+        self._state = None
+        self._dimension = None
+        self._conv_cache = {}
+
+    @property
+    def state(self):
+        return self._state
```
alvarofpp, line 27: If there is no manipulation of _`state`, then this method is unnecessary. Rename `_state` to `state`.

alvarofpp (approved):

urnai/sc2/states/deepmind_state.py
RickFqt: When I tried to do so, some errors ocurred on the tests. There is an abstract method called `state` in the base class `StateBase`, which has a `@property`. I think that if we don't implement this method in `DeepMindState,` using `state` will try to call this abstract method.

